PR #3, opened 2017-04-10
Add Travis CI

danrabbit:


danrabbit:
@tintou FFR don't merge travis branches unless it's already passing :p Travis was disabled here because this was failing :p

Issue #6, opened 2017-04-10
First time you click a link it scrolls to the top [$10]


danrabbit:
The first time you click a link, it scrolls to the top of the conversation widget instead of opening the link. Second time you click, it works okay

PR #9, opened 2017-04-24
Fixed: First time you click a link it scrolls to the top

Philip-Scott:
Fixes: #6


Djaler:
This fix has a regression: now I can't copy text from message by Ctrl+C

Djaler:
You should find a way to keep StylishWebView focusable, but prevent scrolling

Philip-Scott:
ahh shit, it was too good to be true. Good catch!

PR #21, opened 2017-05-14
Fix builds with Travis CI

purejava:
Add cmake option SUPPRESS_C_WARNINGS.
- When enabled with cmake -DSUPPRESS_C_WARNINGS=ON .. the majority of the C compiler warnings are turned off.
- This is needed to allow builds with Travis CI, as the service limits the amount of log messages during compilation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,9 @@ env:
 
 install:
  - docker pull elementary/docker:loki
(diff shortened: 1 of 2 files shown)


purejava:
With this PR, the Travis job log looks like [this](https://travis-ci.org/purejava/mail/builds/232100651).

PR #29, opened 2017-05-27
Add action for adding inline images

### LINKED ISSUE #20: Cannot insert images in mail body
When building an HTML mail, I can use an image as an attached file, but I cannot insert it in the mail body, along with the text.

Neither can I save in my computer, nor download, an embedded image when someone sends me an e-mail with an image inside the mail body.

davidmhewitt:
Fixes #20 .

Again, this is a pretty large diff (sorry!). But only code relating to adding attachments and the encoding of outgoing mail has been changed.

I've tested all the combinations of sending mail with and without attachments in different character sets and all seems to be working fine. Might just need a couple more days of testing before merging.


danrabbit:
As far as I can tell, this works as expected. UX Approve :)

PR #60, opened 2017-07-13
Save last window size

danrabbit:


davidmhewitt:
Is there any reason for not saving the window size/position to GSettings when the window closes rather than every time it changes? Just thinking of those precious CPU cycles that we've got billions of anyway :wink:

danrabbit:
Just if the app crashes :man_shrugging:

tintou:
Yeah I would really prefer to save it on app-close, we assume that the app will never crash because we are the best programmers on earth.

PR #67, opened 2017-07-16
Display relative date time in conversation list

danrabbit:
This duplicates the functionality from https://github.com/elementary/mail/blob/master/src/client/util/util-date.vala

Show messages received:
- [x] Less than a minute ago as "Now"
- [x] Less than an hour ago as "5m ago"
- [x] Today, Less than 12 hours ago as "1h ago"
- [x] All other times Today as "10:04 AM"
    - [x] With localized 12/24 hours
- [x] Yesterday as "Yesterday"
- [x] Less than a week ago as "Sat"
- [x] This year as "Jun 15"
- [x] All other times as "nn/nn/nn", localized

### DIFF
--- a/src/ConversationList/ConversationListItem.vala
+++ b/src/ConversationList/ConversationListItem.vala
@@ -144,9 +144,9 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
             return date_time.format ("%a");
         }
         if (date_time.get_year () == now.get_year ()) {
-            return date_time.format ("%b %e");
+            return date_time.format ("%b %-e");
         }
-        return date_time.format ("%b %e %Y");
+        return date_time.format ("%x");
     }
 
     private bool is_same_day (DateTime day1, DateTime day2) {
--- a/src/ConversationList/ConversationListItem.vala
+++ b/src/ConversationList/ConversationListItem.vala
@@ -122,31 +122,25 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
         if (is_same_day (date_time, now)) {
             if (diff < TimeSpan.MINUTE) {
                 return _("Now");
-            }
-            if (diff < TimeSpan.HOUR) {
-                return ngettext ("%dm ago", "%dm ago", (ulong) (diff / TimeSpan.MINUTE)).printf ((int) (diff / TimeSpan.MINUTE));
-            }
-            if (diff < 12 * TimeSpan.HOUR) {
+            } else if (diff < TimeSpan.HOUR) {
+                var minutes = diff / TimeSpan.MINUTE;
+                return ngettext ("%dm ago", "%dm ago", (ulong) (minutes)).printf ((int) (minutes));
+            } else if (diff < 12 * TimeSpan.HOUR) {
                 int rounded = (int) Math.round ((double) diff / TimeSpan.HOUR);
                 return ngettext ("%dh ago", "%dh ago", (ulong) rounded).printf (rounded);
-            }
-            var settings = new Settings ("org.gnome.desktop.interface");
-            if (settings.get_enum ("clock-format") == 0) {
-                return date_time.format ("%H:%M");
             } else {
-                return date_time.format ("%l:%M %p");
+                var settings = new Settings ("org.gnome.desktop.interface");
+                return Granite.DateTime.get_default_time_format (settings.get_enum ("clock-format") == 1, false);
             }
-        }
-        if (is_same_day (date_time.add_days (1), now)) {
+        } else if (is_same_day (date_time.add_days (1), now)) {
             return _("Yesterday");
-        }
-        if (diff < 6 * TimeSpan.DAY) {
+        } else if (diff < 6 * TimeSpan.DAY) {
             return date_time.format ("%a");
+        } else if (date_time.get_year () == now.get_year ()) {
+            return date_time.format (_("%b %-e"));
+        } else {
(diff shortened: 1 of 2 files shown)


src/ConversationList/ConversationListItem.vala
```diff
@@ -114,4 +114,45 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
...
+            }
+        }
+        if (is_same_day (date_time.add_days (1), now)) {
+            return _("Yesterday");
+        }
+        if (diff < 6 * TimeSpan.DAY) {
+            return date_time.format ("%a");
+        }
+        if (date_time.get_year () == now.get_year ()) {
+            return date_time.format ("%b %-e");
```
tintou: this needs to be localized

src/ConversationList/ConversationListItem.vala
```diff
@@ -114,4 +114,45 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
...
+            if (diff < 12 * TimeSpan.HOUR) {
+                int rounded = (int) Math.round ((double) diff / TimeSpan.HOUR);
+                return ngettext ("%dh ago", "%dh ago", (ulong) rounded).printf (rounded);
+            }
+            var settings = new Settings ("org.gnome.desktop.interface");
+            if (settings.get_enum ("clock-format") == 0) {
+                return date_time.format ("%H:%M");
+            } else {
+                return date_time.format ("%l:%M %p");
+            }
```
tintou: there is a Granite util that takes care of all of this for you

src/ConversationList/ConversationListItem.vala
```diff
@@ -114,4 +114,45 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
...
+    private string get_relative_datetime (DateTime date_time) {
+        var now = new DateTime.now_local ();
+        var diff = now.difference (date_time);
+
+        if (is_same_day (date_time, now)) {
+            if (diff < TimeSpan.MINUTE) {
+                return _("Now");
+            }
+            if (diff < TimeSpan.HOUR) {
+                return ngettext ("%dm ago", "%dm ago", (ulong) (diff / TimeSpan.MINUTE)).printf ((int) (diff / TimeSpan.MINUTE));
```
tintou: Don't divide twice, store the value and reuse it instead

src/ConversationList/ConversationListItem.vala
```diff
@@ -114,4 +114,45 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
...
     }
+
+    private string get_relative_datetime (DateTime date_time) {
+        var now = new DateTime.now_local ();
+        var diff = now.difference (date_time);
+
+        if (is_same_day (date_time, now)) {
+            if (diff < TimeSpan.MINUTE) {
+                return _("Now");
+            }
```
tintou: Use } else if
even if the return doesn't make a difference, it makes if clearer when you read it

src/ConversationList/ConversationListItem.vala
```diff
@@ -114,4 +114,45 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
...
+            if (diff < TimeSpan.HOUR) {
+                return ngettext ("%dm ago", "%dm ago", (ulong) (diff / TimeSpan.MINUTE)).printf ((int) (diff / TimeSpan.MINUTE));
+            }
+            if (diff < 12 * TimeSpan.HOUR) {
+                int rounded = (int) Math.round ((double) diff / TimeSpan.HOUR);
```
tintou: another way to calculate it is:
(diff - (diff % TimeSpan.HOUR)) / TimeSpan.HOUR
diff % TimeSpan.HOUR returns diff - max (X * TimeSpan.HOUR) and thus the division doesn't need to be truncated afterward

src/ConversationList/ConversationListItem.vala
tintou: https://valadoc.org/granite/Granite.DateTime.html

src/ConversationList/ConversationListItem.vala
tintou: also you want to do
`diff - diff % TimeSpan.MINUTE` before to remove all the seconds, and milliseconds parts

src/ConversationList/ConversationListItem.vala
eyelash: @tintou but `Math.round()` does not simply truncate. It rounds to the nearest integer, so 1.9 hours will be rounded to 2.

src/ConversationList/ConversationListItem.vala
eyelash: > also you want to do
> `diff - diff % TimeSpan.MINUTE` before to remove all the seconds, and milliseconds parts

No need for that. TimeSpan is an integral type and integer division will be automatically truncated.

tintou: Right, the first comment remains valid

src/ConversationList/ConversationListItem.vala
tintou: Yeah, which seems wrong to me, we don't want to round but to truncate here

danrabbit: I disagree I think we do want to round. If the actually number of hours is 1.9 and we display 1h we're much more incorrect than if we say 2h

PR #71, opened 2017-07-16
Pass Camel.Folder through to MessageListItem

davidmhewitt:
In order to get the content of an individual message, it seems we need a reference to the folder it came from. Pass the folder reference through so we can get ready to parse the message and display in the WebView.


davidmhewitt:
Closing this as there is a way of getting the `Camel.Folder` via `MessageInfo` so this is overly complicated.

PR #76, opened 2017-07-16
Animate message list

tintou:


danrabbit:
Need that `collapsed` class :p

tintou:
Added

tintou:
Yeah this is not the scope of this merge request, I'm just adding the mechanism here

danrabbit:
When you collapse a message the header it disappears. I also can't confirm the collapsed class is being applied

screenshot from 2017-07-17 13 51 56

PR #77, opened 2017-07-16
Align and set the width of the search bar

danrabbit:
A little bit of magic numbers, but not sure how to help that. I guess theoretically if we could get the allocation size for the titlebutton box we could plug that in, but I'm not sure how we'd go about doing that


davidmhewitt:
It's possible to calculate your magic 50 number from the window like this. Hopefully should be more robust in the various cases where that 50 number would change.

```
diff --git a/src/MainWindow.vala b/src/MainWindow.vala
index 2d9e86c7..552858b5 100644
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -104,8 +104,12 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
             message_list_box.set_conversation (node);
         });
 
-        set_paned_start_grid_width ();
-        set_search_entry_width ();
+        ulong initial_allocate;
+        initial_allocate = headerbar.paned_start_grid.size_allocate.connect (() => {
+            set_search_entry_width ();
+            set_paned_start_grid_width ();
+            headerbar.paned_start_grid.disconnect (initial_allocate);
+        });
 
         paned_end.notify["position"].connect (() => set_search_entry_width ());
 
@@ -126,6 +130,10 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
     }
     
     private void set_paned_start_grid_width () {
-        headerbar.paned_start_grid.width_request = paned_start.position - 50;
+        Gtk.Allocation grid_alloc;
+        headerbar.paned_start_grid.get_allocation (out grid_alloc);
+        Gtk.Allocation header_alloc;
+        headerbar.get_allocation (out header_alloc);
+        headerbar.paned_start_grid.width_request = paned_start.position - (grid_alloc.x - header_alloc.x + headerbar.spacing);
     }
 }

```

src/MainWindow.vala
```diff
@@ -99,10 +104,28 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
...
+        });
+
         Backend.Session.get_default ().start.begin ();
     }
 
     private void on_compose_message () {
         new ComposerWindow ().show_all ();
     }
+
+    private void set_search_entry_width () {
```
donadigo: Can we change this name to something like `update_search_entry_width`? It is confusing when a method name starts with `set` but doesn't take any parameters.

src/MainWindow.vala
```diff
@@ -99,10 +104,28 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
...
+        headerbar.search_entry.width_request = paned_end.position - paned_start.position + 1;
+    }
+    
+    private void set_paned_start_grid_width () {
```
donadigo: The same is here.

src/MainWindow.vala
```diff
@@ -99,10 +104,28 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
             message_list_box.set_conversation (node);
         });
 
+        set_paned_start_grid_width ();
+        set_search_entry_width ();
+
+        paned_end.notify["position"].connect (() => set_search_entry_width ());
```
donadigo: Could be just:
`...connect (set_search_entry_width)`.

src/HeaderBar.vala
```diff
@@ -19,17 +19,27 @@
  */
 
 public class Mail.HeaderBar : Gtk.HeaderBar {
+    public Gtk.Grid paned_start_grid;
```
donadigo: Perhaps add a `get; construct;` here?

src/HeaderBar.vala
```diff
@@ -19,17 +19,27 @@
...
+    public Gtk.SearchEntry search_entry;
```
donadigo: Same here.

src/MainWindow.vala
```diff
@@ -99,10 +104,42 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
...
+        headerbar.search_entry.width_request = paned_end.position - paned_start.position + 1;
+    }
+    
+    private void update_paned_start_grid_width () {
+        int offset = 0;
+        headerbar.forall ((widget) => {
+            if (widget.get_style_context ().has_class ("left")) {
+                Gtk.Allocation alloc;
+                widget.get_allocation (out alloc);
+                offset = alloc.width;
```
davidmhewitt: We should probably add a `return` at the end here to save some more of those CPU cycles. My bad.

src/MainWindow.vala
```diff
@@ -99,10 +104,43 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
...
+                return;
+            }
+        });
+
+        int padding = headerbar.get_style_context ().get_padding (Gtk.StateFlags.NORMAL).left;
+        offset += headerbar.spacing + (padding * 2);
```
eyelash: Why are you adding the padding twice here? For me the search bar is a few pixels off and adding the padding only once fixes that. I'm testing this on GNOME 3.24 on Arch with both the Adwaita and Elementary themes as I couldn't compile it on elementary OS stable. What are you guys using to develop this?

tintou:
@eyelash Can you test this: https://github.com/elementary/mail/pull/111/commits/16c79f168b3e52ca277a97ebbe56335282a52e4a and tell me if it solves your issue ?
We are using Ubuntu 17.10 internally to test it.

eyelash:
@tintou it does not fix the problem. I found the problem, however. It occurs because on my setup there is no button on the left side of the header bar (the close button is on the right side on GNOME). I believe the correct fix is to only add the left padding once and add `headerbar.spacing` twice in case there is a widget with class `"left"`. (note that in my case there is no widget with class `"left"`).

eyelash:
Maybe the following ASCII diagram makes things a bit more clear:
```
 -------------------
| x |     | Search
--------------------
 ^  ^     ^
 |  |    headerbar.spacing
 |  |
 | headerbar.spacing
 |
left header bar padding
```

I don't have much time right now but I'm thinking of something like the following implementation:

```vala
        var style_context = headerbar.get_style_context ();
        int padding = style_context.get_padding (style_context.get_state ()).left;
        
        int offset = padding;
        headerbar.forall ((widget) => {
            if (widget.get_style_context ().has_class ("left")) {
                offset += widget.get_allocated_width ();
                offset += headerbar.spacing;
                return;
            }
        });
        offset += headerbar.spacing;

        headerbar.paned_start_grid.width_request = paned_start.position - offset;
```

PR #81, opened 2017-07-17
Remote image blocking

davidmhewitt:
Add two settings keys, one to always allow image loading (defaults to true), another as an email address whitelist of  sender addresses for which remote images will be automatically loaded.

I've not added any UI for controlling the settings keys, nor have I added the infobar that will show when images are blocked (though I have left a TODO comment on the signal where that would be).


danrabbit:
I can confirm that images are blocked with this branch. Probably want Corentin's approve on the async stuff though :)

PR #83, opened 2017-07-18
Open attachment dialog fixes

danrabbit:
* Rename to resolve ambiguity with future FileDialog subclass
* Move to `Dialogs` as in other elementary apps
* Fix styling, sizing
* Resolve terminal warnings about no transient parent, etc

### DIFF
--- a/src/Dialogs/OpenAttachmentDialog.vala
+++ b/src/Dialogs/OpenAttachmentDialog.vala
@@ -21,11 +21,12 @@
 public class Mail.OpenAttachmentDialog : Gtk.Dialog {
     public Camel.MimePart mime_part { get; construct; }
 
-    public OpenAttachmentDialog (Camel.MimePart mime_part) {
+    public OpenAttachmentDialog (Gtk.Window parent, Camel.MimePart mime_part) {
         Object (
             deletable: false,
             mime_part: mime_part,
-            resizable: false
+            resizable: false,
+            transient_for: parent
         );
     }
 
@@ -68,8 +69,6 @@ public class Mail.OpenAttachmentDialog : Gtk.Dialog {
                 show_file_anyway.begin ();
             }
         });
-
-        show_all ();
     }
 
     private async void show_file_anyway () {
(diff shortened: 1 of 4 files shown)


src/MessageList/AttachmentBar.vala
```diff
@@ -69,8 +69,8 @@ public class Mail.AttachmentBar : Gtk.FlowBox {
     }
 
     private void show_attachment (Camel.MimePart attachment_part) {
-        var dialog = new Mail.AttachmentDialog (attachment_part);
-        dialog.transient_for = get_toplevel () as Gtk.Window;
+        var dialog = new Mail.OpenAttachmentDialog ((Gtk.Window) get_toplevel (), attachment_part);
```
tintou: Here the `as` keyword is important, if it's not a Gtk.Window we don't want to use it.

tintou:
don't forget to change the file name in po/POTFILES

src/Dialogs/OpenAttachmentDialog.vala
```diff
@@ -61,16 +69,14 @@ public class Mail.AttachmentDialog : Gtk.Dialog {
...
-        show_all ();
     }
 
     private async void show_file_anyway () {
         try {
             GLib.FileIOStream iostream;
             var file = File.new_tmp ("XXXXXX-%s".printf (mime_part.get_filename ()), out iostream);
             yield mime_part.content.decode_to_output_stream (iostream.output_stream, GLib.Priority.DEFAULT, null);
-            yield GLib.AppInfo.launch_default_for_uri_async (file.get_uri (), null, null);
+            yield GLib.AppInfo.launch_default_for_uri_async (file.get_uri (), new AppLaunchContext (), null);
```
tintou: The binding here is misleading, it's nullable so If you want to avoid the compilation warning, better do `(GLib.AppLaunchContext) null`

Reference: https://developer.gnome.org/gio/stable/GAppInfo.html#g-app-info-launch-default-for-uri-async

PR #84, opened 2017-07-18
Show images InfoBar

danrabbit:
Probably not the best way to do this, but it does the thing. Maybe @davidmhewitt can make it better since he knows how the web view API works :p

### DIFF
--- a/src/MessageList/MessageListItem.vala
+++ b/src/MessageList/MessageListItem.vala
@@ -101,17 +101,34 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         var separator = new Gtk.Separator (Gtk.Orientation.HORIZONTAL);
         separator.hexpand = true;
 
+        settings = new GLib.Settings ("io.elementary.mail");
+
         var infobar = new Gtk.InfoBar ();
         infobar.margin = 12;
         infobar.message_type = Gtk.MessageType.WARNING;
         infobar.add_button (_("Show Images"), 1);
         infobar.add_button (_("Always Show from Sender"), 2);
-        infobar.get_content_area ().add (new Gtk.Label (_("This message contains remote images.")));
         infobar.get_style_context ().add_class (Gtk.STYLE_CLASS_FRAME);
+        infobar.no_show_all = true;
+
+        var infobar_content = infobar.get_content_area ();
+        infobar_content.add (new Gtk.Label (_("This message contains remote images.")));
+        infobar_content.show_all ();
 
         ((Gtk.Box) infobar.get_action_area ()).orientation = Gtk.Orientation.VERTICAL;
 
-        settings = new GLib.Settings ("io.elementary.mail");
+        infobar.response.connect ((id) => {
+            var sender = message_info.from;
+            var whitelist = settings.get_strv ("remote-images-whitelist");
+            if (id == 2) {
+                if (!(sender in whitelist)) {
+                    whitelist += sender;
+                    settings.set_strv ("remote-images-whitelist", whitelist);
+                }
+            }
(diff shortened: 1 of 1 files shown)


src/MessageList/MessageListItem.vala
```diff
@@ -101,21 +101,27 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+        infobar.get_content_area ().add (new Gtk.Label (_("This message contains remote images.")));
+        infobar.get_style_context ().add_class (Gtk.STYLE_CLASS_FRAME);
+
+        ((Gtk.Box) infobar.get_action_area ()).orientation = Gtk.Orientation.VERTICAL;
+
         settings = new GLib.Settings ("io.elementary.mail");
 
         web_view = new Mail.WebView ();
-        web_view.margin = 6;
-        web_view.image_load_blocked.connect (() => {
```
davidmhewitt: Should probably use this signal to determine when to show the infobar rather than checking the whitelist again.

src/MessageList/MessageListItem.vala
```diff
@@ -124,11 +130,33 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+
+        infobar.response.connect ((id) => {
+            if (id == 2) {
+                if (!(sender in whitelist)) {
+                    whitelist += sender;
+                    settings.set_strv ("remote-images-whitelist", whitelist);
+                }
+            }
+            web_view.load_images ();
+            get_message.begin ();
```
davidmhewitt: Could you remove this line that reloads the message? That's my bad for not implementing the part that loads images on demand yet :disappointed:

src/MessageList/MessageListItem.vala
```diff
@@ -124,11 +130,33 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+            infobar.hide ();
+        });
+
+        if (sender in whitelist) {
+            web_view.load_images ();
```
davidmhewitt: Don't need to do this, the webview already correctly loads images for whitelisted emails.

src/MessageList/MessageListItem.vala
```diff
@@ -124,11 +130,33 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         }
 
         add (base_grid);
-        show_all ();
 
         destroy.connect (() => {
             loading_cancellable.cancel ();
         });
+
+        var sender = message_info.from;
```
davidmhewitt: As discussed in slack, you need to do the following to get just the email address:
```
string from_address;
message.get_from ().@get (0, null, out from_address);
```

Although, you don't have access to the message reference here I don't think.

danrabbit:
I'm gonna mark this as blocked until we:
- [ ] implement image loading in web_view.load_images ();
- [ ] Parse email addresses out, because we have to do this in the header anyways and it looks like that snippet doesn't work here

src/MessageList/MessageListItem.vala
```diff
@@ -103,19 +103,43 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+        infobar.no_show_all = true;
+
+        var infobar_content = infobar.get_content_area ();
+        infobar_content.add (new Gtk.Label (_("This message contains remote images.")));
+        infobar_content.show_all ();
+
+        ((Gtk.Box) infobar.get_action_area ()).orientation = Gtk.Orientation.VERTICAL;
+
+        infobar.response.connect ((id) => {
+            var sender = message_info.from;
```
danrabbit: re-adding a note that this needs to get just the email address without the sender's name

PR #87, opened 2017-07-19
show an icon in the header if the message contains attachments

danrabbit:


src/MessageList/MessageListItem.vala
```diff
@@ -96,7 +96,15 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         header.attach (to_val_label, 2, 1, 1, 1);
         header.attach (subject_val_label, 2, 2, 3, 1);
         header.attach (datetime_label, 3, 0, 1, 1);
-        header.attach (starred_icon, 4, 0, 1, 1);
+        header.attach (starred_icon, 5, 0, 1, 1);
+
+        if (Camel.MessageFlags.ATTACHMENTS in (int) message_info.flags) {
```
davidmhewitt: There's a matching if statement on [Line 130](https://github.com/elementary/mail/blob/497d620bd3b46f63782048e08597732a126fa869/src/MessageList/MessageListItem.vala#L130). Can we merge the two? :slightly_smiling_face:

PR #89, opened 2017-07-19
Prevent focusing the WebView scrolling the message to the top (rewrite)

davidmhewitt:


src/MainWindow.vala
```diff
@@ -46,6 +46,9 @@ public class Mail.MainWindow : Gtk.Window {
         var message_list_scrolled = new Gtk.ScrolledWindow (null, null);
         message_list_scrolled.hscrollbar_policy = Gtk.PolicyType.NEVER;
         message_list_scrolled.add (message_list_box);
+        // Prevent the focus of the webview causing the ScrolledWindow to scroll
+        var viewport = message_list_scrolled.get_child () as Gtk.Container;
```
danrabbit: I think in this case we want to cast like `(Gtk.Container) message_list_scrolled.get_child ()` since using `as` there is a possibility to return `null` right?

PR #91, opened 2017-07-19
Stop clicked links from trying to load in the email browser

davidmhewitt:
Also add a `link_activated` signal ready for opening a browser for links that are clicked.

### DIFF
--- a/src/WebView.vala
+++ b/src/WebView.vala
@@ -22,6 +22,9 @@ public extern const string WEBKIT_EXTENSION_PATH;
 
 public class Mail.WebView : WebKit.WebView {
     public signal void image_load_blocked ();
+    public signal void link_activated (string url);
+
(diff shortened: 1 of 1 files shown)


danrabbit:
Do we need to add a signal or could/should we just open it right there with:
```
        try {
            AppInfo.launch_default_for_uri (nav_policy.request.uri, null);
        } catch (Error e) {
            warning ("%s\n", e.message);
        }
```

davidmhewitt:
@danrabbit This WebView widget will presumably be used in the composer as well, I think we want to edit links on click in the composer and open them in the viewer. So I think a signal works best here as it allows the different views to implement the action in their own way :slightly_smiling_face:

danrabbit:
You are right! Good thinking :)

PR #92, opened 2017-07-19
Open links in a web browser

danrabbit:


src/MessageList/MessageListItem.vala
```diff
@@ -135,6 +135,14 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         destroy.connect (() => {
             loading_cancellable.cancel ();
         });
+
+        web_view.link_activated.connect ((uri) => {
+            try {
+                AppInfo.launch_default_for_uri (uri, null);
+            } catch (Error e) {
+                warning ("%s\n", e.message);
```
davidmhewitt: You don't need a newline in the string, `warning` adds that.
Plus can we have a bit of a message that says what the error is?
"Failed to open link: %s" or something

PR #97, opened 2017-07-21
Make collapsed cards more compact

danrabbit:
Not sure if this should be merged, but putting this here for visibility. Probably needs to be played with for a while to see how it feels in practice. Looks like:

screenshot from 2017-07-20 13 10 30

### DIFF
--- a/data/application.css
+++ b/data/application.css
@@ -1,3 +1,7 @@
+.card.collapsed {
+    margin-bottom: -64px;
+}
+
 .conversation-list-item {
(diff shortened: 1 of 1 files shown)


cassidyjames:
I really like how this looks. I think we can try it out and see how it feels. I like that it keeps the from, timestamp, and favorite status all visible (and vertically centered) while saving some space and also overlapping the avatar, which helps imply there's more content to be revealed.

:+1:

tintou:
I like it :+1:

PR #98, opened 2017-07-21
Expand only-child and last-child emails

danrabbit:


src/MessageList/MessageListBox.vala
```diff
@@ -39,6 +39,23 @@ public class Mail.MessageListBox : Gtk.ListBox {
...
+
+        if (children.length () == 1) {
+            children.foreach ((child) => {
+                if (child is MessageListItem) {
+                    ((MessageListItem) child).expanded = true;
+                }
+            });
+        } else {
+            var child = get_row_at_index ((int) children.length () - 1);
+            if (child == null && !(child is MessageListItem)) {
```
davidmhewitt: Can we invert the logic here and bring the `expanded = true` inside the if statement. I don't like the idea of using a return in case someone doesn't notice it and tags more code on the end of the method.

src/MessageList/MessageListBox.vala
```diff
@@ -39,6 +39,23 @@ public class Mail.MessageListBox : Gtk.ListBox {
         if (node.child != null) {
             go_down ((Camel.FolderThreadNode?) node.child);
         }
+
+        var children = get_children ();
+
+        if (children.length () == 1) {
+            children.foreach ((child) => {
```
davidmhewitt: Not a huge fan of using a foreach for a single element list. Can you use `nth_data` or `get_row_at_index` again instead?

PR #101, opened 2017-07-22
Add preferences menu

danrabbit:
Depends on https://github.com/elementary/stylesheet/pull/153 to be styled correctly

screenshot from 2017-07-22 11 32 49

There's only one real preference at this time, so no dialog. If/when there are more preferences we can do a dialog, but if there's only a couple settings this is just faster and easier, imo. If we end up having more of these switch menu items, I'll refactor that into a widget class

Account settings are all external, so assume that the system online accounts settings knows what to do.

### DIFF
--- a/src/HeaderBar.vala
+++ b/src/HeaderBar.vala
@@ -29,7 +29,52 @@ public class Mail.HeaderBar : Gtk.HeaderBar {
         compose_button.action_name = "win." + MainWindow.ACTION_COMPOSE_MESSAGE;
 
         var search_entry = new Gtk.SearchEntry ();
+
+        var load_images_switch = new Gtk.Switch ();
+
+        var load_images_grid = new Gtk.Grid ();
+        load_images_grid.column_spacing = 6;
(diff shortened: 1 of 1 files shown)


src/HeaderBar.vala
```diff
@@ -39,7 +39,59 @@ public class Mail.HeaderBar : Gtk.HeaderBar {
...
+        load_images_grid.margin_end = 6;
+        load_images_grid.margin_start = 6;
+        load_images_grid.add (new Gtk.Label (_("Always Show Remote Images")));
+        load_images_grid.add (load_images_switch);
+
+        var load_images_menuitem = new Gtk.Button ();
+        load_images_menuitem.get_style_context ().add_class (Gtk.STYLE_CLASS_MENUITEM);
+        load_images_menuitem.add (load_images_grid);
+
+        var account_settings_menuitem = new Gtk.ModelButton ();
```
davidmhewitt: I don't think it's a problem, but why use a `ModelButton` instead of a `MenuItem` here?

PR #102, opened 2017-07-22
Add composer tool buttons

davidmhewitt:


src/Composer/ComposerWindow.vala
```diff
@@ -52,17 +52,77 @@ public class Mail.ComposerWindow : Gtk.ApplicationWindow {
...
+
+        var image = new Gtk.Button.from_icon_name ("insert-image-symbolic", Gtk.IconSize.MENU);
+        image.margin_top = image.margin_left = 6;
+        image.tooltip_text = _("Image (Ctrl+G)");
+
+        var clear_format = new Gtk.Button.from_icon_name ("format-text-clear-formatting-symbolic", Gtk.IconSize.MENU);
+        clear_format.margin_top = clear_format.margin_left = 6;
+        clear_format.tooltip_text = _("Remove formatting (Ctrl+Space)");
+
+        var button_box = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 0);
```
danrabbit: Instead of adding margin top and left to each of these, probably should make this a Gtk.Grid and use `column_spacing` and add a top margin to the grid, no?

PR #112, opened 2017-07-23
improve the offset calculation

eyelash:
See https://github.com/elementary/mail/pull/77#issuecomment-317276078 for an explanation.

You should be able to test this by running
`gsettings set org.gnome.desktop.wm.preferences button-layout :close`

Also, the offset is now calculated from left to right by first adding the padding.


tintou:
You might want to look at https://github.com/elementary/mail/pull/113
It's a attempt to make it really layout independent

danrabbit:
Closing since this was superseded by #113

PR #113, opened 2017-07-23
Layout independant and robust spacing widget size calculation

tintou:


src/HeaderBar.vala
```diff
@@ -19,26 +19,27 @@
...
-    public Gtk.Grid paned_start_grid { get; construct; }
     public Gtk.SearchEntry search_entry { get; construct; }
+    private Gtk.Grid spacing_widget;
 
     public HeaderBar () {
         Object (show_close_button: true);
     }
 
     construct {
+        custom_title = new Gtk.Grid ();
```
danrabbit: This doesn't seem to get used anywhere

tintou: It is required because Gtk creates a fake title widget with no way to discriminate it, so we are overriding it with a simple one.

PR #115, opened 2017-07-24
Add first 4 font formatting actions to composer

davidmhewitt:


tintou:
This looks good, I'm wondering if we should maybe create an abstract composer view that handles all the signal so that we can reuse it in the window composer and the embedded composer

davidmhewitt:
@tintou I've moved all of the code that can be re-used in both composers into a separate class. It's still not ideal because of the ActionBar being below the WebView which won't be suitable for the inline composer. But it's a lot more separated. Let me know if there's anything you'd do differently.

davidmhewitt:
This is now implemented in the `ComposerWidget` class which is used for both the windowed and inline composer. The same action code can therefore be used in both.

PR #120, opened 2017-07-28
Add inline composer

davidmhewitt:


src/Composer/InlineComposer.vala
```diff
@@ -0,0 +1,49 @@
...
+
+        get_style_context ().add_class ("card");
+
+        var composer = new ComposerWidget ();
+        composer.margin_top = 6;
+        composer.has_recipients = true;
+
+        var content_grid = new Gtk.Grid ();
+        content_grid.orientation = Gtk.Orientation.VERTICAL;
+        content_grid.add (composer);
```
danrabbit: Is this grid necessary? Seems like you only pack one thing into it :p

src/HeaderBar.vala
```diff
@@ -78,9 +78,15 @@ public class Mail.HeaderBar : Gtk.HeaderBar {
         app_menu.popover = app_menu_popover;
         app_menu.tooltip_text = _("Menu");
 
+        var reply_button = new Gtk.Button.from_icon_name ("mail-reply-sender", Gtk.IconSize.LARGE_TOOLBAR);
+        reply_button.halign = Gtk.Align.START;
```
danrabbit: I don't think it's necessary to add an halign  here since hexpand is false by default

src/MessageList/MessageListBox.vala
```diff
@@ -66,4 +66,8 @@ public class Mail.MessageListBox : Gtk.ListBox {
             current_node = (Camel.FolderThreadNode?) current_node.child;
         }
     }
+
+    public void reply () {
+        prepend (new InlineComposer ());
```
danrabbit: shouldn't this be an add and then we scroll to bottom since new messages will appear below old messages?

PR #121, opened 2017-07-30
Add remove formatting action

davidmhewitt:


src/Composer/ComposerWidget.vala
```diff
@@ -175,6 +178,14 @@ public class Mail.ComposerWidget : Gtk.Grid {
         web_view.query_command_state (command);
     }
 
+    private void on_remove_format () {
+        web_view.execute_editor_command ("removeformat");
+        web_view.execute_editor_command ("removeparaformat");
+        web_view.execute_editor_command ("unlink");
+        web_view.execute_editor_command ("backcolor", "#ffffff");
+        web_view.execute_editor_command ("forecolor", "#000000");
```
tintou: Shouldn't removeformat already cover the backcolor and forecolor ?
Because that doesn't feel like "removing" things to set a color…

davidmhewitt: You're right. I've copied this method from Geary because I didn't want to change the behavior too much. But after I've sat and read the spec a little more, it would seem that the `removeformat` command at least removes the foreground color and the `removeparaformat` command doesn't even exist in WebKit. So maybe I should trust this Geary code a little less :wink:

PR #122, opened 2017-07-31
Handle Folder changes, refresh message list

tintou:


src/ConversationList/ConversationListBox.vala
```diff
@@ -48,26 +51,71 @@ public class Mail.ConversationListBox : Gtk.ListBox {
...
+                }
+            });
 
-        cancellable = new GLib.Cancellable ();
-        try {
-            Camel.Folder? folder = yield ((Camel.Store) current_account.service).get_folder (current_folder, Camel.StoreGetFolderFlags.BODY_INDEX, GLib.Priority.DEFAULT, cancellable);
-            /*yield folder.refresh_info (GLib.Priority.DEFAULT, cancellable);*/
-            thread = new Camel.FolderThread (folder, folder.get_uids (), false);
             unowned Camel.FolderThreadNode? child = (Camel.FolderThreadNode?) thread.tree;
             while (child != null) {
```
davidmhewitt: Is this iterating over the whole folder again? The `FolderChangeInfo`  class has `get_added_uids` and `get_changed_uids` which feel like a more efficient way to handle the changes.

src/FoldersView/FoldersListView.vala
```diff
@@ -34,16 +34,9 @@ public class Mail.FoldersListView : Gtk.ScrolledWindow {
...
-        foreach (var account in session.get_accounts ()) {
-            var account_item = new Mail.AccountSourceItem (account);
-            source_list.root.add (account_item);
-        }
-
-        session.account_added.connect ((account) => {
-            var account_item = new Mail.AccountSourceItem (account);
-            source_list.root.add (account_item);
-        });
+        session.get_accounts ().foreach ((account) => {add_account (account); return true;});
```
davidmhewitt: I think I'd like to see this broken out into a separate `foreach` statement again, or at least onto more than one line. Just feels like there's a lot squeezed onto one line.

src/ConversationList/ConversationListBox.vala
tintou: The problem is that `get_added_uids` doesn't give us in which thread it is added, so we need to iterate the threadnodes again…

PR #123, opened 2017-08-01
Add basic email autocompletion

tintou:


danrabbit:
I'm getting `(io.elementary.mail:4334): eds-WARNING **: edsf-persona-store.vala:2609: Error in address book view query: Authentication required`

danrabbit:
After making Evolution grab the contacts I can confirm that it works :)

danrabbit:
I think the avatar here is probably too small to be useful.

screenshot from 2017-08-01 13 25 00

btkostner:
I like the idea of having two lines. One for the name, and one for the email.

PR #127, opened 2017-08-02
Sort starred messages to the top of the list

danrabbit:
Following up on https://bugs.launchpad.net/pantheon-mail/+bug/1590976

Not so sure about this anymore for a couple of reasons:

1. There's a starred folder in the sidebar so these messages are already easy to get to
2. The listbox lazy loads so it won't show all starred messages from ever, only recent ones

### DIFF
--- a/src/ConversationList/ConversationListBox.vala
+++ b/src/ConversationList/ConversationListBox.vala
@@ -119,6 +119,18 @@ public class Mail.ConversationListBox : Gtk.ListBox {
     private static int thread_sort_function (Gtk.ListBoxRow row1, Gtk.ListBoxRow row2) {
         var item1 = (ConversationListItem) row1;
         var item2 = (ConversationListItem) row2;
-        return (int)(item2.node.message.date_received - item1.node.message.date_received);
+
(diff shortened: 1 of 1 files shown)


davidmhewitt:
The starred folder only exists if your mailserver has one. I feel like that's a gmail thing.

danrabbit:
I guess we could enforce that there is always a starred folder if we don't find one?

tintou:
Camel supports virtual folders so it's possible to create one if there is none

danrabbit:
Yeah I kinda am leaning towards enforcing the folder to be the better solution here. That way we're not cluttering up the inbox (Inbox 0 is the mission amirite?) and there's still an easy way to get to these messages

danrabbit:
Gonna close this since it sounds like there's a general consensus that this isn't a great workflow for everyone

cassidyjames:
I **really** like the starred-first approach (and use it in FastMail for my work email), but I agree that it might not be a generic enough thing to do by default. Perhaps down the road it could return as an option or something (having them sorted up top makes them unignorable, which is nice for Inbox Zero). :man_shrugging:

PR #130, opened 2017-08-07
Fix an oops

davidmhewitt:
Was getting quite a few `GLib-GIO-ERROR **: Settings schema 'io.elementary.mail' does not contain a key named 'window-maximized'` errors. `maximized` has been used throughout the code, missing a `d` in the schema.


danrabbit:
Hey uh wait. This is against master and org.pantheon.mail but your description says io.elementary.mail and you've tagged rewrite. Which is it? :p

PR #136, opened 2017-08-11
Make the WebKit extension the server

davidmhewitt:
Previously, the UI process of Mail was the DBus server and each instance of the WebKit extension connected to it. To get information out of the WebKit DOM, we had to fire a signal on the UI side that the extension listened to and then the extension sent something back to the server.

This was a bit of a pain and didn't really make sense. So, now the webkit extension is the DBus server and the UI connects to it and asks it to do stuff.

Bit of a nasty diff since all the logic had to be reversed, but it's more red :wink: 

For testing, all functionality should be exactly the same.


danrabbit:
Looks like the web view is no longer editable in this branch. Oops :p

PR #141, opened 2017-08-12
Add the To and CC fields to the inline composer

davidmhewitt:
Fields don''t get auto populated in the inline composer yet. Just moving the fields into the parent class and hiding the subject box in the inline composer.

### DIFF
--- a/src/Composer/ComposerWidget.vala
+++ b/src/Composer/ComposerWidget.vala
@@ -57,8 +57,12 @@ public class Mail.ComposerWidget : Gtk.Grid {
         }
     }
 
-    public ComposerWidget (bool show_subject_field = true) {
-        Object (has_subject_field: show_subject_field);
+    public ComposerWidget () {
+        Object (has_subject_field: false);
+    }
+
+    public ComposerWidget.with_subject () {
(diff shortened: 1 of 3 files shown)


src/Composer/ComposerWidget.vala
```diff
@@ -51,19 +53,96 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
             _has_recipients = value;
-            update_send_sensitivity ();
+            send.sensitive = has_recipients;
         }
     }
 
-    private void update_send_sensitivity () {
-        send.sensitive = has_recipients;
+    public ComposerWidget (bool show_subject_field = true) {
+        Object (has_subject_field: show_subject_field);
```
danrabbit: Since this is currently a construction property, would it be totally overkill to have:

```
public ComposerWidget () {
    Object (has_subject_field: false);
}

public ComposerWidget.with_subect () {
    Object (has_subject_field: true:);
}
```

I feel like that would be a little less "mystery meat" looking from the outside when you're doing `new ComposerWidget (true);` vs `new ComposerWidget.with_subject ()`

PR #201, opened 2018-03-29
EDS rewrite: Minor display fixes

jlnr:
Two small fixes:

* When opening an empty folder, the previous conversation remains selected (unlike on the non-EDS master build). I've made the selected conversation an optional parameter, and null indicates no selection.
* I've sent myself a few mails using Evolution and they have no `date_received`. I've changed the code to fall back on the `date_sent` field in this case.

### DIFF
--- a/src/ConversationList/ConversationListItem.vala
+++ b/src/ConversationList/ConversationListItem.vala
@@ -30,6 +30,16 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
 
     public Camel.FolderThreadNode node { get; private set; }
 
+    public int64 timestamp {
+        get {
+            if (node.message.date_received == 0) {
+                // Sent messages do not have a date_received timestamp.
+                return node.message.date_sent;
+            }
+            return node.message.date_received;
+        }
+    }
+
     public ConversationListItem (Camel.FolderThreadNode node) {
         update_node (node);
     }
@@ -108,12 +118,7 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
             flagged_icon_revealer.reveal_child = true;
         }
 
-        var relevant_timestamp = node.message.date_received;
-        if (relevant_timestamp == 0) {
-            // Sent messages do not have a date_received timestamp.
-            relevant_timestamp = node.message.date_sent;
-        }
-        date.label = Granite.DateTime.get_relative_datetime (new DateTime.from_unix_local (relevant_timestamp));
+        date.label = Granite.DateTime.get_relative_datetime (new DateTime.from_unix_local (timestamp));
     }
 
     private static uint count_thread_messages (Camel.FolderThreadNode node) {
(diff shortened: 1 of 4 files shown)


src/MessageList/MessageListBox.vala
```diff
@@ -30,26 +30,27 @@ public class Mail.MessageListBox : Gtk.ListBox {
...
+            return;
+
         var item = new MessageListItem (node.message);
         add (item);
         if (node.child != null) {
             go_down ((Camel.FolderThreadNode?) node.child);
         }
 
         var children = get_children ();
-        if (children.length () == 1) {
```
jlnr: This special case looked identical to the general case, so I've removed it

danrabbit:
Hey thanks for your branch! Someone should be able to review soon. In the meantime, I sent you an invite to our Slack. I'd really love to get work going on this branch again :)

src/MessageList/MessageListBox.vala
```diff
@@ -30,26 +30,27 @@ public class Mail.MessageListBox : Gtk.ListBox {
...
-            var child = get_row_at_index (0);
-            if (child is MessageListItem) {
-                var list_item = (MessageListItem) child;
-                list_item.expanded = true;
-                list_item.bind_property ("loaded", this, "can-reply", BindingFlags.SYNC_CREATE);
-            }
-        } else {
+        if (children.length () > 0) {
```
tintou: I know that this was the case before but as we are using many time `children.length ()` it might be a good thing to cache this value as getting the length of an array can be an expensive operation

src/MessageList/MessageListItem.vala
```diff
@@ -139,7 +139,12 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         header_stack.add_named (small_fields_grid, "small");
         header_stack.show_all ();
 
-        var datetime_label = new Gtk.Label (new DateTime.from_unix_utc (message_info.date_received).format ("%b %e, %Y"));
+        var relevant_timestamp = message_info.date_received;
+        if (relevant_timestamp == 0) {
+            // Sent messages do not have a date_received timestamp.
+            relevant_timestamp = message_info.date_sent;
+        }
```
tintou: Small codestyle issue, put an empty (as in no indentation) new line between end of curly bracket and new code

src/ConversationList/ConversationListItem.vala
```diff
@@ -30,6 +30,16 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
 
     public Camel.FolderThreadNode node { get; private set; }
 
+    public int64 timestamp {
+        get {
+            if (node.message.date_received == 0) {
+                // Sent messages do not have a date_received timestamp.
+                return node.message.date_sent;
+            }
```
tintou: Small codestyle issue, put an empty (as in no indentation) new line between end of curly bracket and new code

jlnr:
@tintou 👋 Thanks for the review! I've fixed the newlines, but I wasn't sure what the most idiomatic name for the length variable would be. And should I use `array.length` or `array.length ()`?
The variable is only used twice (no loop), can `length` really be this slow?

jlnr:
@tintou Oh I see, I was getting arrays and lists confused. I'll update it in the next PR.

PR #202, opened 2018-03-30
EDS rewrite: Correct encoding, display plain text emails

jlnr:
I couldn't read plain text mails in my inbox and tracked it down to two issues:

* The messages were ISO-8859 encoded, and WebView.load_plain_text fails on invalid UTF-8 strings. I have fixed this by converting message contents to UTF-8 where they are loaded.
* WebView.load_plain_text was not overridden to set the load_changed handler like WebView.load_html. I've added an implementation using Camel's text-to-HTML filter.

Re: encoding stuff, it feels odd to have so much model logic in MessageListItem. Should I move the asynchronous message decoding to a helper class?

### DIFF
--- a/src/MessageList/MessageListItem.vala
+++ b/src/MessageList/MessageListItem.vala
@@ -389,6 +389,16 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
             }
 
             message_content = (string) os.steal_data ();
+
+            if (!message_content.validate ()) {
+                // If message_content is not valid UTF-8, assume that it is
+                // ISO-8859-1 encoded and convert it to UTF-8.
+                // This prevents silent rendering failures down the line because
+                // WebKit's load_plain_text can only handle valid UTF-8 strings.
+                // TODO: This should use the correct encoding from the Content-Type header.
+                message_content = GLib.convert (message_content, -1, "UTF-8", "ISO-8859-1");
+            }
+
             if (field.subtype == "html") {
                 message_is_html = true;
             }
--- a/src/MessageList/MessageListItem.vala
+++ b/src/MessageList/MessageListItem.vala
@@ -388,21 +388,44 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
                 return;
             }
 
-            message_content = (string) os.steal_data ();
-
-            if (!message_content.validate ()) {
-                // If message_content is not valid UTF-8, assume that it is
-                // ISO-8859-1 encoded and convert it to UTF-8.
-                // This prevents silent rendering failures down the line because
-                // WebKit's load_plain_text can only handle valid UTF-8 strings.
-                // TODO: This should use the correct encoding from the Content-Type header.
-                message_content = GLib.convert (message_content, -1, "UTF-8", "ISO-8859-1");
-            }
+            // Convert the message to UTF-8 to ensure we have a valid GLib string.
+            message_content = convert_to_utf8(os, field.param ("charset"));
 
             if (field.subtype == "html") {
                 message_is_html = true;
             }
         }
+    }
+
+    private static string convert_to_utf8(GLib.MemoryOutputStream os, string? encoding) {
+    	var num_bytes = (int) os.get_data_size ();
+		var bytes = (string) os.steal_data ();
+
+		string? result = null;
+
+        if (encoding != null) {
+            string? iconv_encoding = Camel.iconv_charset_name (encoding);
+            if (iconv_encoding != null) {
+                try {
+                    result = GLib.convert (bytes, num_bytes, "UTF-8", iconv_encoding);
+                } catch (ConvertError e) {
+                    // Nothing to do - result will be assigned below.
+                }
+            }
+        }
+
+        if (result == null || !result.validate()) {
+            // If message_content is not valid UTF-8 at this point, assume that
+            // it is ISO-8859-1 encoded by default, and convert it to UTF-8.
+            try {
+                warning ("No encoding/charset found for email; using ISO-8859-1");
+                result = GLib.convert (bytes, num_bytes, "UTF-8", "ISO-8859-1");
+            } catch (ConvertError e) {
+                critical("Every string should be valid ISO-8859-1");
+            }
+        }
+
+        return result;
     }
 
     private async void handle_inline_mime (Camel.MimePart part) {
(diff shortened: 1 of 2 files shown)


src/WebView.vala
```diff
@@ -160,6 +160,21 @@ public class Mail.WebView : WebKit.WebView {
         }
     }
 
+    public new void load_plain_text (string plain_text) {
```
tintou: Please do not use the `new` keyboard to hide the parent class method but either:
 * create a new method with a different name
 * if `load_plain_text` it's only used once, replace it with the content of the code.

Explanation:
the new keyboard allow you to use the same method name as one of the parent ones but it has some hidden consequences in the code:
If you have the webview somewhere in the code declared as `Webkit.WebView` and that you call `load_plain_text` then it won't call this method but the one in `Webkit.WebView`
I think that the behavior you want is the one used with the `override` keyboard when it would actually replace the original function for all calls. But as this method is not `virtual` not `abstract` you can't use the `override` keyword.

src/WebView.vala
```diff
@@ -160,6 +160,21 @@ public class Mail.WebView : WebKit.WebView {
...
+        // Instead of calling base.load_plain_text, use Camel's ToHTML filter to
+        // convert text to HTML. This gives us some niceties like clickable URLs
+        // and email addresses.
+
+        // Explanation of ToHTML flags:
+        // https://wiki.gnome.org/Apps/Evolution/Camel.MimeFilter#Camel.MimeFilterToHtml
```
tintou: Please use multiline comment style:
```
/*
 * Instead of calling base.load_plain_text, use Camel's ToHTML filter to
 * convert text to HTML. This gives us some niceties like clickable URLs
 * and email addresses.
 *
 * Explanation of ToHTML flags:
 * https://wiki.gnome.org/Apps/Evolution/Camel.MimeFilter#Camel.MimeFilterToHtml
 */```

File: src/MessageList/MessageListItem.vala [hunk 3]
```diff
@@ -388,13 +388,45 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+                    utf8 = GLib.convert (bytes, num_bytes, "UTF-8", iconv_encoding);
+                } catch (ConvertError e) {
+                    // Nothing to do - result will be assigned below.
+                }
+            }
+        }
+
+        if (utf8 == null || !utf8.validate()) {
+            // If message_content is not valid UTF-8 at this point, assume that
+            // it is ISO-8859-1 encoded by default, and convert it to UTF-8.
```
tintou: Please use multiline comment style:
```
/* 
 * If message_content is not valid UTF-8 at this point, assume that
 * it is ISO-8859-1 encoded by default, and convert it to UTF-8.
 */```

File: src/MessageList/MessageListItem.vala [hunk 4]
```diff
@@ -388,13 +388,45 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
                 return;
             }
 
-            message_content = (string) os.steal_data ();
+            // Convert the message to UTF-8 to ensure we have a valid GLib string.
+            message_content = convert_to_utf8(os, field.param ("charset"));
```
tintou: Missing a space between function name and parenthesis opening

src/MessageList/MessageListItem.vala
```diff
@@ -388,13 +388,45 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+            }
+        }
+
+        if (utf8 == null || !utf8.validate()) {
+            // If message_content is not valid UTF-8 at this point, assume that
+            // it is ISO-8859-1 encoded by default, and convert it to UTF-8.
+            try {
+                utf8 = GLib.convert (bytes, num_bytes, "UTF-8", "ISO-8859-1");
+            } catch (ConvertError e) {
+                critical("Every string should be valid ISO-8859-1");
```
tintou: Missing a space between function name and parenthesis opening
You can also include the e.message text in the critical message so that we have clear explanation of why the conversion failed.
You also probably want to return an UTF-8 encoded HTML page containing an `unable to load message` error message and the explanation :man_shrugging:

src/MessageList/MessageListItem.vala
```diff
@@ -388,13 +388,45 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+            if (iconv_encoding != null) {
+                try {
+                    utf8 = GLib.convert (bytes, num_bytes, "UTF-8", iconv_encoding);
+                } catch (ConvertError e) {
+                    // Nothing to do - result will be assigned below.
+                }
+            }
+        }
+
+        if (utf8 == null || !utf8.validate()) {
```
tintou: missing a space between `validate` and the parenthesis

src/MessageList/MessageListItem.vala
```diff
@@ -388,13 +388,45 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+            // Convert the message to UTF-8 to ensure we have a valid GLib string.
+            message_content = convert_to_utf8(os, field.param ("charset"));
+
             if (field.subtype == "html") {
                 message_is_html = true;
             }
         }
     }
 
+    private static string convert_to_utf8(GLib.MemoryOutputStream os, string? encoding) {
```
tintou: missing space between `convert_to_utf8` and the parenthesis

src/WebView.vala
jlnr: Hmmm, do you think we should rename `new void load_html (...)` to `void set_html (...)` too them? I thought reusing the `load_plain_text` name would be more consistent with `load_html`.

Since `load_plain_text` is only called once, I've inlined the code there (in MessageListItem). -> One more thing that can move into a utility class later :)

src/MessageList/MessageListItem.vala
jlnr: Ouch, forgot to check for code style in today's commits. I should probably write a Builder plugin :)

src/MessageList/MessageListItem.vala
jlnr: That's a good idea about the error page, but I think I'd prefer to postpone this for now. It would require localization work, and I *think* that this can't happen anyway.

src/MessageList/MessageListItem.vala
```diff
@@ -353,13 +353,26 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+                Camel.MimeFilterToHTMLFlags.CONVERT_ADDRESSES;
+            var html = Camel.text_to_html (message_content, flags, 0);
+            web_view.load_html (html);
         }
     }
 
-    private async void parse_mime_content (Camel.DataWrapper message_content) {
-        if (message_content is Camel.Multipart) {
-            var content = message_content as Camel.Multipart;
+    private async void parse_mime_content (Camel.DataWrapper mime_content) {
```
jlnr: I've renamed this because having both an instance variable and a parameter called `message_content` is a bit confusing

PR #235, opened 2018-08-21
[Rewrite WIP] Port ConversationList to TreeView

### LINKED ISSUE #231: Loading a folder with a lot of emails causes the app to hang
This is due to using a ListBox to render the conversation list. The inefficiencies of creating one widget per email are showing a bit.

For example, there is a slight noticeable delay when opening a folder with approx. 500 emails. My gmail account with approx. 9000 emails hangs the app for so long that I can't be bothered waiting for it to load anymore.

I have tried modifying the conversation list to use a ListStore and only bind it to the ListBox when it has been populated. This offers a small speedup but still nowhere near large enough to be usable. I'd say people that keep a lot of emails in a folder is a pretty common use case, so it looks like this should be re-written as a TreeView again, despite the downsides of using a TreeView.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61903568-loading-a-folder-with-a-lot-of-emails-causes-the-app-to-hang?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
**This is still under heavy development.** Raising a PR for visibility that I'm working on the issue.

Fixes #231 

Still to do:
- [ ] Sort TreeView by date again
- [ ] Add pixbuf to denote unread messages
- [ ] Add hoverable?/clickable pixbuf for starred/flagged state
- [ ] Hook up the open message signals again
- [ ] Check we can still update individual messages/threads in real-time rather than having to reload the treeview
- [ ] Test thoroughly, especially with RTL layouts and both HiDPI and non-HiDPI displays


danrabbit:
I would really prefer not to go back to using treeview here if possible. Maybe a simpler solution is to just load the first 20 emails and only load more emails on `Gtk.ScrolledWindow.edge_reached`?

davidmhewitt:
Sorting and filtering/searching become much more problematic when you do that, as does having a scrollbar that actually works properly. There are also a bunch of other issues around memory management. If you scroll really far down what do you do with all the widgets that are off the top of the display, do you keep those around or recycle the RAM?

I can see the downsides of using a treeview but the amount of work in developing a custom widget that handles all of the above correctly is much more of an undertaking than just porting this to treeview. Granted, that custom widget would probably be useful in a bunch of other places (hell, even in gtk itself). But, let me know if treeview is a definite no and I'll abandon this branch.

tintou:
I'm unsure about this as Gtk.Treeview is considered as deprecated by the Gtk+ maintainers themselves (although there is currently no efficient replacement for it)

davidmhewitt:
@tintou Understood. I've been working on a more efficient ListBox implementation over the past few days and will submit a new PR for that soon. Though I'll keep this open for now in case that doesn't work out either.

cassidyjames:
@davidmhewitt considering the conflicts here, do you want to resolve them and keep this open, close it, or just keep it around for reference?

davidmhewitt:
I'll close it, but keep the branch around for now in case I need to refer back to it.

PR #237, opened 2018-08-31
Rewrite: Use newest timestamp in thread for thread timestamp

davidmhewitt:
In multi-message threads, it seems like the first message in the thread was being used to calculate the timestamp of that particular thread.

This PR takes whichever is the newest timestamp in the thread and uses that for the entire thread instead ensuring that threads with newer messages are sorted to the top.

### DIFF
--- a/src/ConversationList/ConversationListItem.vala
+++ b/src/ConversationList/ConversationListItem.vala
@@ -32,12 +32,36 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
 
     public int64 timestamp {
         get {
+            int64 timestamp = 0;
             if (node.message.date_received == 0) {
                 // Sent messages do not have a date_received timestamp.
                 return node.message.date_sent;
+                timestamp = node.message.date_sent;
+
+                unowned Camel.FolderThreadNode? child = (Camel.FolderThreadNode?) node.child;
+                while (child != null) {
+                    if (child.message.date_sent > timestamp) {
+                        timestamp = child.message.date_sent;
+                    }
+
+                    child = (Camel.FolderThreadNode?) child.next;
+                }
+
+                return timestamp;
             }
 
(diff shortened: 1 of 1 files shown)


src/ConversationList/ConversationListItem.vala
```diff
@@ -32,12 +32,36 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
 
     public int64 timestamp {
         get {
+            int64 timestamp = 0;
             if (node.message.date_received == 0) {
                 // Sent messages do not have a date_received timestamp.
                 return node.message.date_sent;
+                timestamp = node.message.date_sent;
```
jlnr: This will not be reached because the `return` statement on the previous line is still there.

src/ConversationList/ConversationListItem.vala
```diff
@@ -32,12 +32,36 @@ public class Mail.ConversationListItem : Gtk.ListBoxRow {
...
+                    child = (Camel.FolderThreadNode?) child.next;
+                }
+
+                return timestamp;
             }
 
             return node.message.date_received;
+            timestamp = node.message.date_received;
+            unowned Camel.FolderThreadNode? child = (Camel.FolderThreadNode?) node.child;
+            while (child != null) {
```
jlnr: Instead of branching on whether `node.message.data_received != 0`, and then looping over all messages in the message thread, I think we should have a single loop through the message thread, and then calculate the `max(date_received, date_sent)` (pseudocode) of all messages.
Otherwise, if the thread starts with a sent message (= no `date_received`), the `date_received` timestamps of other messages in the thread will be ignored.

davidmhewitt:
@jlnr Thanks for the review! Hadn't spotted those extra returns during merging this branch, thank you!

But yes, using your method makes more sense. I've implemented the changes.

PR #265, opened 2018-10-17
Add Insert Link Dialog

### LINKED ISSUE #197: Insert URL dialog's default text is redundant
The dialog inputs the following text by default:
screenshot from 2018-03-18 14-41-23
On opening the dialog box, the text doesn't get selected automatically so that the user can paste the copied URL from the clipboard and have the text replaced.

The user is much more likely to paste an existing copied URL from the clipboard rather than typing the URL in the dialog box, which is why 
1. Either remove the default text
2. Or auto-select it on pop up so that it gets replaced on a paste action from clipboard 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56325045-insert-url-dialog-s-default-text-is-redundant?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #197 

screenshot from 2018-10-17 15 03 52 2x


davidmhewitt:
This is all hooked up now, although we don't have any way of checking where the links go because the composer has no tooltips or actions hooked up for clicking/hovering links.

src/Composer/ComposerWidget.vala
```diff
@@ -296,6 +299,26 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+
+    private void on_link_inserted (string url, string title) {
+        var selected_text = web_view.get_selected_text ();
+        if (selected_text != null && title == selected_text) {
+            web_view.execute_editor_command ("createLink", url);
+        } else {
+            if (title != null && title.length > 0) {
+                web_view.execute_editor_command ("insertHTML", """<a href="%s">%s</a>""".printf (url, title));
+            } else {
+                web_view.execute_editor_command ("insertHTML", """<a href="%s">%s</a>""".printf (url, url));
```
tintou:
```suggestion
                web_view.execute_editor_command ("insertHTML", "<a>%s</a>".printf (url));
```

tintou: Just a tiny thing

davidmhewitt: @tintou I don't think that does what I'd like it to do. An anchor without a href attribute is a placeholder link. Placeholder links are a bit like doing `<a href="#">whatever</a>` (i.e. they don't go anywhere and are designed to be used in Javascript or whatever). I want this to be a real link that goes places, just with the same text as the URL.

PR #271, opened 2018-10-18
Settings: Make window position an array instead of 2 keys

danrabbit:


danrabbit:
@cassidyjames was going for a small diff. I can do a follow up branch for that?

PR #292, opened 2018-10-23
Settings: Use an array for window size

danrabbit:
Since we always change these together


tintou:
I think that it would make more sense to handle it as a Tuple of integers `(ii)`

PR #294, opened 2018-10-23
Add a welcome screen

danrabbit:
Adds a welcome screen that links to Online Accounts.

Need to add a signal listener to show/hide the appropriate screen depending on if accounts are configured in EDS.

screenshot from 2018-10-22 21 36 36 2x


cassidyjames:
Can probably drop "Email" from the top line since it should be pretty obvious from context. Otherwise I can't think of better copy at the moment. :)

PR #295, opened 2018-10-23
Set tooltip accel markup

danrabbit:
Uses the new tooltip markup style that we discussed this morning in #UX 

screenshot from 2018-10-23 10 13 26 2x

I've temporarily added two utils that were recently merged into granite [here](https://github.com/elementary/granite/commit/07afa7ac0cbdf1d6dd923d5392722c54665af0c5). I figure we can easily remove those utils once a new Granite release comes out.

### DIFF
--- a/src/Utils.vala
+++ b/src/Utils.vala
@@ -61,4 +61,64 @@ public class Mail.Utils {
     public static string escape_html_tags (string input) {
         return input.replace ("<", "&lt;").replace (">", "&gt;");
     }
+
+    private static string accel_to_string (string accel) {
+        uint accel_key;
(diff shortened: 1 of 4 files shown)


src/Utils.vala
```diff
@@ -61,4 +61,64 @@ public class Mail.Utils {
     public static string escape_html_tags (string input) {
         return input.replace ("<", "&lt;").replace (">", "&gt;");
     }
+
+    private static string accel_to_string (string accel) {
```
cassidyjames: @danrabbit maybe add a comment that this should be removed and replaced with the Granite util once it's released? It'd be nice to not lose track of it. :smile:

PR #300, opened 2018-10-23
MainWindow.vala: add action accels

danrabbit:


tintou:
I think that you need to chain-up the constructor:
```public MainWindow (Gtk.Application application) {
base (application);
}```
I haven't tested it though

**Comment by danrabbit** [comment]
Looks like no:
```
 chain up to `Gtk.ApplicationWindow.new' not supported
```

**Comment by davidmhewitt** [comment]
`application` in the Gtk.Window class is not being set to the application we pass until after the `construct` block has finished, so it's just a null pointer when you're trying to set accelerators on it.

The accelerator for loop either needs moving to below `Object` in the non-GObject constructor, or, we need to keep our own reference to the application instead/as well.

File: src/MainWindow.vala [hunk 1]
```diff
@@ -53,9 +56,22 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
...
+        action_accelerators[ACTION_FORWARD] = "<Ctrl><Shift>F";
+        action_accelerators[ACTION_MOVE_TO_TRASH] = "Delete";
+        action_accelerators[ACTION_MOVE_TO_TRASH] = "BackSpace";
+    }
+
     construct {
         add_action_entries (action_entries, this);
 
+        foreach (var action in action_accelerators.get_keys ()) {
+            ((Gtk.Application) GLib.Application.get_default ()).set_accels_for_action (ACTION_PREFIX + action, action_accelerators[action].to_array ());
```
jeremypw: Over long line?  I would split in two using a temporary var for the application.

cassidyjames: We've also just broken the arguments over multiple lines in other cases:

```suggestion
            ((Gtk.Application) GLib.Application.get_default ()).set_accels_for_action (
                ACTION_PREFIX + action, 
                action_accelerators[action].to_array ()
            );
```

jeremypw: Is there a downside to a temp variable? I suspect the resulting binary code is much the same anyway if the compiler is optimising.  No big deal either way.

cassidyjames: @jeremypw not that I know of, just giving an alternate suggestion.

PR #336, opened 2018-12-14
Using symbolic icons for the toolbar on Desktop environments other than Pantheon

GabMus:
Fixing a very simple issue, many other desktop environments use symbolic icons for their toolbars, and other than for style coherence, this is also for accessibility. When using symbolic icons, they are visible on the headerbar regardless of the fact that said headerbar is dark or light.

Here are some screenshots showing my change: https://imgur.com/a/PAPSjBZ

### DIFF
--- a/src/HeaderBar.vala
+++ b/src/HeaderBar.vala
@@ -28,9 +28,18 @@ public class Mail.HeaderBar : Gtk.HeaderBar {
     }
 
     construct {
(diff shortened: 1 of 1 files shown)


danrabbit:
Closing as "Out of Scope". Stylesheets can already specify if icons in a selector should be symbolic or color if the platform wants to enforce that convention

PR #349, opened 2019-02-08
ComposerWidget: Add an attachment action

danrabbit:
Todo:
- [x] Make attachment widget into a class so its easy to get the GLib.File for all attachments
- [x] Remove attachments
- [x] ~~Fall back to `text-x-preview` when `get_icon` returns a missing image https://github.com/elementary/icons/pull/778~~
- [ ] Don't let the user try to send and warn when file size is too large. It looks like the smallest limit is from  outlook.com/hotmail which is 10Mb
- [ ] Don't allow adding duplicate attachments

Something to think about is what we do when a user adds an attachment and then deletes the file from their disk. Should we watch the file and just remove it or should we check the files exist before sending and throw an error or something else?

screenshot from 2019-02-10 15 43 03 2x


tintou:
After playing with it, I think that we probably need to move this into a popover, because when you add several attachments (some can have long names too) it takes a big amount of space

src/Composer/ComposerWidget.vala
```diff
@@ -608,14 +639,61 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
         part.set_encoding (Camel.TransferEncoding.ENCODING_QUOTEDPRINTABLE);
+
+        var body = new Camel.Multipart ();
+        body.set_mime_type ("multipart/alternative");
+        body.set_boundary (null);
         body.add_part (part);
+
+        if (attachment_box.get_children ().length () > 0) {
+            foreach (unowned Gtk.Widget attachment in attachment_box.get_children ()) {
+                if (attachment is Attachment) {
```
tintou, line 653:
```suggestion
                if (!(attachment is Attachment)) {
                    continue;
                }
```
This would allow to de-indent it a bit

src/Composer/ComposerWidget.vala
```diff
@@ -608,14 +639,61 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+                    var attachment_obj = (Attachment)attachment;
```
tintou, line 654:
```suggestion
                    unowned var attachment_obj = (Attachment)attachment;
```

src/Composer/ComposerWidget.vala
```diff
@@ -608,14 +639,61 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+                    var file = File.new_for_uri (attachment_obj.uri);
+
+                    var content_type = attachment_obj.info.get_content_type ();
```
tintou, line 657:
```suggestion
                    unowned string? content_type = attachment_obj.info.get_content_type ();
```

src/Composer/ComposerWidget.vala
```diff
@@ -631,6 +709,54 @@ public class Mail.ComposerWidget : Gtk.Grid {
         from_combo.active = 0;
     }
 
+    private class Attachment : Gtk.FlowBoxChild {
+        public GLib.FileInfo info { get; construct; }
+        public string uri { get; construct; }
```
tintou, line 714: Why not keep a GLib.File here ? You seem to get them and recreate them afterward
```suggestion
        public GLib.File file { get; construct; }
```

src/Composer/ComposerWidget.vala
```diff
@@ -405,6 +415,34 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+            _("Choose a file"),
+            (Gtk.Window) get_toplevel (),
+            Gtk.FileChooserAction.OPEN,
+            _("Attach"),
+            _("Cancel")
+        );
+
+        if (filechooser.run () == Gtk.ResponseType.ACCEPT) {
+            filechooser.hide ();
+            foreach (File file in filechooser.get_files ()) {
```
tintou, line 429:
```suggestion
            foreach (unowned File file in filechooser.get_files ()) {
```

src/Composer/ComposerWidget.vala
```diff
@@ -405,6 +415,34 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+                try {
+                    var info = file.query_info ("standard::*", 0);
```
tintou, line 431:
```suggestion
                    var info = file.query_info ("standard::*", 0);
```
I'd prefer that step to be inside the `Attachment` object

tintou: That way we can query less properties than `"standard::*"` which means a lot of things

danrabbit:
Attachment sent, so fuckin' yolo dude :rocket:

PR #352, opened 2019-02-11
Add Mark menu

danrabbit:
Could probably be broken down into individual menu items if this is gonna get complicated

- [ ] Make the menubutton insensitive if there is no selection
- [ ] Mark a thread as spam
- [x] Mark a thread read/unread
- [x] Star a thread
- [ ] Change to "Mark as Read" if the thread is unread
- [ ] Change to "Mark as Not Spam" if the thread is spam
- [ ] Change to "Unstar" if the thread is starred


danrabbit:
Closing since this is superseded by other branches

PR #354, opened 2019-02-13
Add mark read/unread actions and shortcuts

danrabbit:


src/MainWindow.vala
```diff
@@ -46,6 +48,8 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
         {ACTION_REPLY,              on_reply             },
         {ACTION_REPLY_ALL,          on_reply_all         },
         {ACTION_FORWARD,            on_forward           },
+        {ACTION_MARK_READ, on_mark_read },
+        {ACTION_MARK_UNREAD, on_mark_unread },
```
tintou: can you keep this align as above

danrabbit: Yes, but I feel like we should probably not try to do this kind of alignment since it just causes more line changes in the future if we ever update this

PR #355, opened 2019-02-13
MessageListItem: Star a message by clicking the star icon

### LINKED ISSUE #259: Ability to star a message
<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/64895388-ability-to-star-a-message?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #259

### DIFF
--- a/src/MessageList/MessageListItem.vala
+++ b/src/MessageList/MessageListItem.vala
@@ -169,7 +169,7 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         action_grid.halign = Gtk.Align.END;
         action_grid.valign = Gtk.Align.START;
         action_grid.add (datetime_label);
-        action_grid.add (starred_button);
+        action_grid.attach (starred_button, 2, 0);
 
         var header = new Gtk.Grid ();
         header.margin = 12;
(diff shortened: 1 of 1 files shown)


src/MessageList/MessageListItem.vala
```diff
@@ -248,13 +257,26 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
             expanded = !expanded;
-            return false;
+            return Gdk.EVENT_STOP;
         });
 
         destroy.connect (() => {
             loading_cancellable.cancel ();
         });
 
+        starred_button.button_release_event.connect (() => {
```
tintou: Why `button_release_event` instead of `clicked` ?

danrabbit: So we can stop the event from propogating to the eventbox that collapses the mail

tintou: That would be perfect to not change it in the future

src/MessageList/MessageListItem.vala
```diff
@@ -248,13 +257,26 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+            if (Camel.MessageFlags.FLAGGED in (int) message_info.flags) {
+                message_info.set_flags (Camel.MessageFlags.FLAGGED, 0);
+                starred_icon.icon_name = "non-starred-symbolic";
+                starred_icon.tooltip_text = _("Star message");
+            } else {
+                message_info.set_flags (Camel.MessageFlags.FLAGGED, ~0);
```
naaando: What does ~0 means? 🤔

tintou: It means complement all bits of 0, so it's all bits to 1.
With the `Camel.MessageFlags.FLAGGED` mask it means that only the `Camel.MessageFlags.FLAGGED` flag will be set to 1

PR #362, opened 2019-02-14
MessageListItem: Parse message_info.from into a linkbutton

danrabbit:
Replaces the dumb label with a link style button that opens a mailto link, just like Mail 1.x did

### DIFF
--- a/src/MessageList/MessageListItem.vala
+++ b/src/MessageList/MessageListItem.vala
@@ -88,9 +88,7 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         subject_label.valign = Gtk.Align.START;
         subject_label.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
 
-        var from_val_label = new Gtk.Label (message_info.from);
-        from_val_label.wrap = true;
-        from_val_label.xalign = 0;
+        var from_individualitem = new IndividualItem (message_info.from);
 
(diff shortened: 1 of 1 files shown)


tintou:
I would prefer a single Gtk.Grid subclass for the field that holds the buttons and the entry:
Use [Camel.InternetAddress](https://valadoc.org/camel-1.2/Camel.InternetAddress.html) class to parse the addresses instead of splitting `<` and `>` yourself so that it is handled consistently through the system.

src/MessageList/MessageListItem.vala
```diff
@@ -462,4 +460,41 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
     public string get_message_body_html () {
         return web_view.get_body_html ();
     }
+
+    private class IndividualItem : Gtk.Button {
```
tintou: Why not subclass Gtk.LinkButton directly ?

danrabbit:
Ah hey I was looking for something like that, but couldn't find it. I'll update accordingly

danrabbit:
@tintou hm so I'm doing:

```
        string parsed_address;
        string parsed_name;

        var internet_address = new Camel.InternetAddress ();
        internet_address.unformat (message_info.from);
        internet_address.get (0, out parsed_name, out parsed_address);
```

and getting a segfault in malloc. Any ideas?

PR #372, opened 2019-03-10
Full screen support

roymckenzie:
Hello! I liked how Terminal, Code, Epiphany supported true full screen and wanted to have the same with Mail. I use full screen on many of my applications so it's easy to switch using workspaces and have it feel like the entire workspace is dedicated to just that app.

I am not sure if the HIG says that the `HeaderBar` should not be visible when in full screen mode (I noticed it is hidden on Code and Epihany), but core functionality was hidden away so I made the `HeaderBar` visible in full screen mode in my PR.

Full screen Mail

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -22,6 +22,7 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
     private HeaderBar headerbar;
     private Gtk.Paned paned_end;
     private Gtk.Paned paned_start;
+    private Gtk.Grid container_grid;
 
     private FoldersListView folders_list_view;
     private Gtk.Overlay conversation_list_overlay;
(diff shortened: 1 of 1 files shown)


src/MainWindow.vala
```diff
@@ -245,6 +251,20 @@ public class Mail.MainWindow : Gtk.ApplicationWindow {
         }
     }
 
+    private void on_fullscreen () {
+		if ((get_window ().get_state () & Gdk.WindowState.FULLSCREEN) != 0) {
```
tintou: You can use Vala syntax for flags :slightly_smiling_face: 
```suggestion
		if (Gdk.WindowState.FULLSCREEN in get_window ().get_state ()) {
```

roymckenzie:
Thanks!

PR #380, opened 2019-06-04
ComposerWidget: Show a warning when there is no subject

danrabbit:
Screenshot from 2019-06-03 23 11 17@2x

Just kind of bullshitting the secondary text. Could use some help there

### DIFF
--- a/src/Composer/ComposerWidget.vala
+++ b/src/Composer/ComposerWidget.vala
@@ -457,6 +457,27 @@ public class Mail.ComposerWidget : Gtk.Grid {
     }
 
     private void on_send () {
+        if (subject_val.text == "") {
+            var no_subject_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                _("Send Message With an Empty Subject?"),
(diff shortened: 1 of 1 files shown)


src/Composer/ComposerWidget.vala
```diff
@@ -457,6 +457,27 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
 
     private void on_send () {
+        if (subject_val.text == "") {
+            var no_subject_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                _("Send Message With an Empty Subject?"),
+                _("This message has an empty subject field. The recipient may not be able to infer the scope or importance of the message."),
+                "mail-send",
+                Gtk.ButtonsType.NONE
+            );
+            no_subject_dialog.transient_for = (Gtk.Window) get_toplevel ();
```
tintou: We are not sure that it is always a Gtk.Window so let's have this so it's either a Gtk.Window or null
```suggestion
            no_subject_dialog.transient_for = get_toplevel () as Gtk.Window;
```

Issue #357, opened 2019-02-13
mailto links don't open the composer [$20]


danrabbit:
using something like the email referral link from Amazon, Mail opens but the composer doesn't open pre-filled like it should

PR #382, opened 2019-06-06
Open mailto: URLs given as program arguments (#357)

vertexian:
Please let me know if I'm missing any conventions of the project, stylistic or otherwise. I've implemented subject, cc, and recipient, but not body (which is less used and didn't seem worth the trouble of wrangling the WebView).

Also, I waive my claim to the bounty.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -18,9 +18,6 @@
  * Authored by: Corentin Noël <corentin@elementary.io>
  */
 
-using Soup;
-using Gee;
-
 public class Mail.Application : Gtk.Application {
     public static GLib.Settings settings;
 
@@ -58,11 +55,11 @@ public class Mail.Application : Gtk.Application {
 
         // The only arguments we support are mailto: URLs passed in by the OS. See RFC 2368 for
         // details. We handle the most commonly used fields.
-        foreach (var mailtoUri in argv[1:argv.length]) {
+        foreach (var mailto_uri in argv[1:argv.length]) {
             string to = null, cc = null, subject = null;
 
             try {
-                URI mailto = new URI(mailtoUri);
+                Soup.URI mailto = new Soup.URI(mailto_uri);
                 if (mailto == null) {
                     throw new OptionError.BAD_VALUE ("Argument is not a URL.");
                 }
@@ -70,7 +67,7 @@ public class Mail.Application : Gtk.Application {
                     throw new OptionError.BAD_VALUE ("Cannot open non-mailto: URL");
                 }
 
-                to = URI.decode(mailto.path);
+                to = Soup.URI.decode(mailto.path);
                 if (to == null || to == "") {
                     throw new OptionError.BAD_VALUE ("mailto: URL does not specify a recipent");
                 }
@@ -87,15 +84,15 @@ public class Mail.Application : Gtk.Application {
                 new ComposerWindow.with_headers (main_window, to, cc, subject).show_all ();
             }
             catch (OptionError e) {
-                warning("Argument parsing error. %s", e.message);
+                warning ("Argument parsing error. %s", e.message);
             }
         }
 
         return 0;
     }
 
-    private HashMap<string, string> parse_mailto_query (string query) throws OptionError {
-        var result = new HashMap<string, string> ();
+    private Gee.HashMap<string, string> parse_mailto_query (string query) throws OptionError {
+        var result = new Gee.HashMap<string, string> ();
         var params = query.split ("&");
 
         foreach (string param in params) {
@@ -104,7 +101,7 @@ public class Mail.Application : Gtk.Application {
                 throw new OptionError.BAD_VALUE ("Invalid mailto URL");
             }
 
-            result[terms[0]] = URI.decode (terms[1]);
+            result[terms[0]] = Soup.URI.decode (terms[1]);
         }
 
         return result;
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -18,6 +18,9 @@
  * Authored by: Corentin Noël <corentin@elementary.io>
  */
 
+using Soup;
+using Gee;
+
 public class Mail.Application : Gtk.Application {
     public static GLib.Settings settings;
 
@@ -26,7 +29,7 @@ public class Mail.Application : Gtk.Application {
     public Application () {
         Object (
             application_id: "io.elementary.mail",
-            flags: ApplicationFlags.HANDLES_OPEN
+            flags: ApplicationFlags.HANDLES_COMMAND_LINE
         );
     }
 
@@ -48,8 +51,63 @@ public class Mail.Application : Gtk.Application {
         set_accels_for_action ("app.quit", {"<Control>q"});
     }
 
-    public override void open (File[] files, string hint) {
-        activate ();
+    public override int command_line (ApplicationCommandLine command_line) {
+        activate();
+
+        string[] argv = command_line.get_arguments ();
+
+        // The only arguments we support are mailto: URLs passed in by the OS. See RFC 2368 for
+        // details. We handle the most commonly used fields.
+        foreach (var mailtoUri in argv[1:argv.length]) {
+            string to = null, cc = null, subject = null;
+
+            try {
+                URI mailto = new URI(mailtoUri);
+                if (mailto == null) {
+                    throw new OptionError.BAD_VALUE ("Argument is not a URL.");
+                }
+                if (mailto.scheme != "mailto") {
+                    throw new OptionError.BAD_VALUE ("Cannot open non-mailto: URL");
+                }
+
+                to = URI.decode(mailto.path);
+                if (to == null || to == "") {
+                    throw new OptionError.BAD_VALUE ("mailto: URL does not specify a recipent");
+                }
+                if (mailto.query != null) {
+                    var params = parse_mailto_query (mailto.query);
+                    if (params["cc"] != null) {
+                        cc = params["cc"];
+                    }
+                    if (params["subject"] != null) {
+                        subject = params["subject"];
+                    }
+                }
+
+                new ComposerWindow.with_headers (main_window, to, cc, subject).show_all ();
+            }
+            catch (OptionError e) {
+                warning("Argument parsing error. %s", e.message);
+            }
+        }
+
+        return 0;
+    }
+
+    private HashMap<string, string> parse_mailto_query (string query) throws OptionError {
+        var result = new HashMap<string, string> ();
+        var params = query.split ("&");
+
+        foreach (string param in params) {
+            var terms = param.split ("=");
+            if (terms.length != 2) {
+                throw new OptionError.BAD_VALUE ("Invalid mailto URL");
+            }
+
+            result[terms[0]] = URI.decode (terms[1]);
+        }
+
+        return result;
     }
 
     public override void activate () {
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -18,6 +18,9 @@
  * Authored by: Corentin Noël <corentin@elementary.io>
  */
 
+using Soup;
+using Gee;
+
 public class Mail.Application : Gtk.Application {
     public static GLib.Settings settings;
 
@@ -55,11 +58,11 @@ public class Mail.Application : Gtk.Application {
 
         // The only arguments we support are mailto: URLs passed in by the OS. See RFC 2368 for
         // details. We handle the most commonly used fields.
-        foreach (var mailto_uri in argv[1:argv.length]) {
+        foreach (var mailtoUri in argv[1:argv.length]) {
             string to = null, cc = null, subject = null;
 
             try {
-                Soup.URI mailto = new Soup.URI(mailto_uri);
+                URI mailto = new URI(mailtoUri);
                 if (mailto == null) {
                     throw new OptionError.BAD_VALUE ("Argument is not a URL.");
                 }
@@ -67,7 +70,7 @@ public class Mail.Application : Gtk.Application {
                     throw new OptionError.BAD_VALUE ("Cannot open non-mailto: URL");
                 }
 
-                to = Soup.URI.decode(mailto.path);
+                to = URI.decode(mailto.path);
                 if (to == null || to == "") {
                     throw new OptionError.BAD_VALUE ("mailto: URL does not specify a recipent");
                 }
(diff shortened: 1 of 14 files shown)


src/Composer/ComposerWidget.vala
```diff
@@ -77,11 +78,29 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+    public ComposerWidget.from_request ( ComposeRequest req ) {
+        Object (has_subject_field: true, template: req);
+    }
+
     construct {
         actions = new SimpleActionGroup ();
         actions.add_action_entries (action_entries, this);
         insert_action_group (ACTION_GROUP_PREFIX, actions);
 
+        var recipient_grid = construct_header_fields ();
```
vertexian: Note that I've done a refactor of the `construct` block here. I've re-arranged the code into four different methods, each of which returns its top-level element, more or less. This does make it a little hard to find the mailto related changes, but if you grep for `template` you'll see them.

tintou:
Thanks for this, it's starting to look nice!
A few things to fix:
 * you seem to add a libsoup dependency and only use it for URL parsing. The Camel library that we are already including here also handles this so it would probably be better to use this function instead ( https://valadoc.org/camel-1.2/Camel.URL.html )
 * I'm not really a fan of having a class just to hold this information, and you can probably parse everything into the `command_line` function. You can then simply have `public ComposerWidget.from_request (string? to = null, string? subject = null, string? cc = null);`
 * If you happen in the future to refactor the original code, it is a good thing to keep it in a separate commit so that we are sure that the refactor has been done well and that the actual addition is right :)

Nice work !

vertexian:
Thanks for the quick response! And apologies in advance for pushing back a little, as a new contributor :-)

> A few things to fix:
> 
>     * you seem to add a libsoup dependency and only use it for URL parsing. The Camel library that we are already including here also handles this so it would probably be better to use this function instead ( https://valadoc.org/camel-1.2/Camel.URL.html )

So it looks like we're already using Soup instead of Camel in several other places that we do URI parsing. And Camel.URL does this annoying thing where URI decoding mutates the string in place. So I think maybe my mistake was just in trying to add libsoup as a dependency (which I didn't even do correctly). It seems to already be accessible somehow.

>     * I'm not really a fan of having a class just to hold this information, and you can probably parse everything into the `command_line` function. You can then simply have `public ComposerWidget.from_request (string? to = null, string? subject = null, string? cc = null);`

My concern with this is that if I pass the individual fields into the constructor, then I think I still need to store them on ComposerWidget until the `construct` block runs. That is, I don't think I can pass them through directly to the` construct` code? In which case, I think it's nice to have them grouped together.

>     * If you happen in the future to refactor the original code, it is a good thing to keep it in a separate commit so that we are sure that the refactor has been done well and that the actual addition is right :)

Agreed. I definitely should have done it this way.

tintou:
> Thanks for the quick response! And apologies in advance for pushing back a little, as a new contributor :-)
> 
> > A few things to fix:
> > you seem to add a libsoup dependency and only use it for URL parsing. The Camel library that we are already including here also handles this so it would probably be better to use this function instead ( https://valadoc.org/camel-1.2/Camel.URL.html )
> 
> So it looks like we're already using Soup instead of Camel in several other places that we do URI parsing. And Camel.URL does this annoying thing where URI decoding mutates the string in place. So I think maybe my mistake was just in trying to add libsoup as a dependency (which I didn't even do correctly). It seems to already be accessible somehow.
It is probably a subdependency of Camel or Webkit2Gtk4
I don't understand how it "mutates" the string ? Can you be more specific ?
> 
> > I'm not really a fan of having a class just to hold this information, and you can probably parse everything into the `command_line` function. You can then simply have `public ComposerWidget.from_request (string? to = null, string? subject = null, string? cc = null);`
> 
> My concern with this is that if I pass the individual fields into the constructor, then I think I still need to store them on ComposerWidget until the `construct` block runs. That is, I don't think I can pass them through directly to the` construct` code? In which case, I think it's nice to have them grouped together.
The object creating goes like this:
```vala
public ComposerWidget (string? arg) {
    //Here the widget doesn't exist
    Object (property1: value1, property2: value2);
    // Here the `construct` has already been called so we can call any method and access members
    set_arg (arg);
}
```
> 
> > If you happen in the future to refactor the original code, it is a good thing to keep it in a separate commit so that we are sure that the refactor has been done well and that the actual addition is right :)
> 
> Agreed. I definitely should have done it this way.

vertexian:
> > So it looks like we're already using Soup instead of Camel in several other places that we do URI parsing. And Camel.URL does this annoying thing where URI decoding mutates the string in place. So I think maybe my mistake was just in trying to add libsoup as a dependency (which I didn't even do correctly). It seems to already be accessible somehow.
> 
> It is probably a subdependency of Camel or Webkit2Gtk4
> I don't understand how it "mutates" the string ? Can you be more specific ?
https://valadoc.org/camel-1.2/Camel.URL.decode.html

> The object creating goes like this:
> 
> ```vala
> public ComposerWidget (string? arg) {
>     //Here the widget doesn't exist
>     Object (property1: value1, property2: value2);
>     // Here the `construct` has already been called so we can call any method and access members
>     set_arg (arg);
> }
> ```
Aha. For some reason I was thinking that the `construct` block ran after all the regular constructors. I'll go ahead and implement as you've suggested then.

src/Composer/ComposerWidget.vala
danrabbit: I would really like to avoid using methods for this. It doesn't feel very object-oriented to me. If it feels cleaner to not have these widgets constructed here then we should probably create classes instead of methods

src/Composer/ComposerWindow.vala
```diff
@@ -19,18 +19,27 @@
  */
 
 public class Mail.ComposerWindow : Gtk.ApplicationWindow {
+    public ComposeRequest template { get; construct; }
```
danrabbit: If possible, I think it helps to make variable names include the class name. So for example, `compose_request` or `template_composerequest` makes it more clear that this is an instance of `ComposeRequest`. This really helps when someone else later down the line is trying to debug the code and wondering what kind of object `template` is

src/Composer/ComposerWindow.vala
```diff
@@ -19,18 +19,27 @@
...
+
     public ComposerWindow (Gtk.Window parent) {
+        this.from_request (parent, null);
```
danrabbit: elementary code style avoids `this` where possible. I realize that isn't in the code style guide. Sorry, our mistake!

src/Composer/ComposerWidget.vala
vertexian: I do believe that the existing constructor method is way too long for a single function, and adding classes is overcomplicating things. Functional deconstruction seems the cleanest approach. That said, I noticed you have a pending PR that this would create a messy merge conflict with, so best to save this refactor for later.

src/Composer/ComposerWindow.vala
vertexian: So, I don't think it's avoidable in this instance (calling one constructor from another)? I tried a couple other syntaxes, and they didn't work.

vertexian:
Refactored to not use a separate class for the mailto parsing logic, and to avoid the refactor of the `ComposerWidget` constructor for now. I've kept the use of `Soup.URI` because, as noted earlier, it's what we use elsewhere for URI parsing.

src/Application.vala
```diff
@@ -18,6 +18,9 @@
  * Authored by: Corentin Noël <corentin@elementary.io>
  */
 
+using Soup;
```
danrabbit: elementary codestyle is to explicitly namespace. So instead of `using foo`, we make sure to disambiguate with `foo.bar` each time. This is mainly because some libraries use the same class names for different things and it can cause issues when you're unsure if this is a `Gtk.Foo` or ` GLib.Foo` or maybe a `Granite.Foo` or `Clutter.Foo`, etc etc

src/Application.vala
```diff
@@ -48,8 +51,63 @@ public class Mail.Application : Gtk.Application {
...
-    public override void open (File[] files, string hint) {
-        activate ();
+    public override int command_line (ApplicationCommandLine command_line) {
+        activate();
+
+        string[] argv = command_line.get_arguments ();
+
+        // The only arguments we support are mailto: URLs passed in by the OS. See RFC 2368 for
+        // details. We handle the most commonly used fields.
+        foreach (var mailtoUri in argv[1:argv.length]) {
```
danrabbit: elementary code style is for variables to use `snake_case` instead of `camelCase`

src/Application.vala
```diff
@@ -48,8 +51,63 @@ public class Mail.Application : Gtk.Application {
...
+                    }
+                    if (params["subject"] != null) {
+                        subject = params["subject"];
+                    }
+                }
+
+                new ComposerWindow.with_headers (main_window, to, cc, subject).show_all ();
+            }
+            catch (OptionError e) {
+                warning("Argument parsing error. %s", e.message);
```
danrabbit: elementary code style is spaces before `()`, so this should be:
```suggestion
                warning ("Argument parsing error. %s", e.message);
```

src/Composer/ComposerWindow.vala
```diff
@@ -19,18 +19,21 @@
...
         Object (
             height_request: 600,
             title: _("New Message"),
             transient_for: parent,
             width_request: 680,
             window_position: Gtk.WindowPosition.CENTER_ON_PARENT
         );
-    }
 
-    construct {
```
danrabbit: We want to keep with using GObject-style construction here, so if you want to pass variables into the `construct` block from a constructor, you need to create public `{ get; construct; default = null}` properties for `to`, `cc`, and `subject`.

To make each constructor more sane and non-repeating, we probably should take this time to move anything that isn't set in the constructor arguments into the `construct` block like the `width_request`, `window_position`, etc.

src/Application.vala
vertexian: Ha, was trying to do that, but old habits die hard.

src/Composer/ComposerWindow.vala
vertexian: I'll implemented the rest of the suggestions, but this I'd like to hang on to. My initial implementation used a separate class to group these fields and encapsulate the parsing. I was told not to do it this way, and to create a constructor with the fields split out. And now I feel like I'd be adding in extra verbosity of new properties, for something that's purely as stylistic convention. (I don't think folding the construct block into the primary constructor changes the semantics.) I think it's okay to break convention sometimes for the sake of simplicity.

I'm happy to go back to the separate ComposeRequest class, but at some point this sort of constant rewriting of PRs is too much work to be putting in for free. I need to be trusted to be able to make a certain amount of design choices of my own.

vertexian:
Backed out some cruft that I'd accidentally commited.

src/Composer/ComposerWindow.vala
danrabbit: I understand if this is beyond the scope of the work you intended to put in. I've submitted a branch [here](https://github.com/vertexian/mail/pull/1) to fix this issue

vertexian:
Merged @danrabbit's constructor style fix. Should be ready for re-review.

src/Application.vala
```diff
@@ -48,8 +48,63 @@ public class Mail.Application : Gtk.Application {
         set_accels_for_action ("app.quit", {"<Control>q"});
     }
 
-    public override void open (File[] files, string hint) {
-        activate ();
+    public override int command_line (ApplicationCommandLine command_line) {
+        activate();
```
tintou:
```suggestion
        activate ();
```

src/Application.vala
```diff
@@ -48,8 +48,63 @@ public class Mail.Application : Gtk.Application {
...
+
+        string[] argv = command_line.get_arguments ();
+
+        // The only arguments we support are mailto: URLs passed in by the OS. See RFC 2368 for
+        // details. We handle the most commonly used fields.
+        foreach (var mailto_uri in argv[1:argv.length]) {
+            string to = null, cc = null, subject = null;
+
+            try {
+                Soup.URI mailto = new Soup.URI(mailto_uri);
```
tintou:
```suggestion
                Soup.URI mailto = new Soup.URI (mailto_uri);
```

src/Application.vala
```diff
@@ -48,8 +48,63 @@ public class Mail.Application : Gtk.Application {
...
+                if (mailto == null) {
+                    throw new OptionError.BAD_VALUE ("Argument is not a URL.");
+                }
+                if (mailto.scheme != "mailto") {
+                    throw new OptionError.BAD_VALUE ("Cannot open non-mailto: URL");
+                }
+
+                to = Soup.URI.decode(mailto.path);
```
tintou:
```suggestion
                to = Soup.URI.decode (mailto.path);
```

src/Application.vala
```diff
@@ -48,8 +48,63 @@ public class Mail.Application : Gtk.Application {
...
+                    if (params["cc"] != null) {
+                        cc = params["cc"];
+                    }
+                    if (params["subject"] != null) {
+                        subject = params["subject"];
+                    }
+                }
+
+                new ComposerWindow.with_headers (main_window, to, cc, subject).show_all ();
+            }
```
tintou: no newline here

src/Application.vala
```diff
@@ -48,8 +48,63 @@ public class Mail.Application : Gtk.Application {
...
+        }
+
+        return 0;
+    }
+
+    private Gee.HashMap<string, string> parse_mailto_query (string query) throws OptionError {
+        var result = new Gee.HashMap<string, string> ();
+        var params = query.split ("&");
+
+        foreach (string param in params) {
```
tintou:
```suggestion
        foreach (unowned string param in params) {
```

src/Composer/ComposerWidget.vala
```diff
@@ -77,6 +78,20 @@ public class Mail.ComposerWidget : Gtk.Grid {
         Object (has_subject_field: true);
     }
 
+    public ComposerWidget.with_headers ( string? to, string? cc, string? subject ) {
```
tintou:
```suggestion
    public ComposerWidget.with_headers (string? to, string? cc, string? subject) {
```

src/Composer/ComposerWidget.vala
```diff
@@ -77,6 +78,20 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+        Object (has_subject_field: true);
+
+        // Set header fields, now that signal handlers are established
+        if (to != null)
+            to_val.text = to;
+        if (cc != null) {
+            cc_button.clicked();
```
tintou:
```suggestion
            cc_button.clicked ();
```

src/Composer/ComposerWidget.vala
```diff
@@ -77,6 +78,20 @@ public class Mail.ComposerWidget : Gtk.Grid {
         Object (has_subject_field: true);
     }
 
+    public ComposerWidget.with_headers ( string? to, string? cc, string? subject ) {
+        Object (has_subject_field: true);
+
+        // Set header fields, now that signal handlers are established
+        if (to != null)
```
tintou: We always use braces, even for single line if:
```vala
if (to != null) {
    to_val.text = to;
}
```

src/Application.vala
vertexian: I'll ask about this in #vala, because I think I understand why this is necessary now, but it seems subtle.

danrabbit:
Fixes #357

PR #388, opened 2019-06-08
Remove unused po files

cjfloss:
Well, most of them.

```bash
cd mail/po
# Remove all duplicates (Empty translations are all the same)
fdupes -NdS . | grep '[+]' | sed 's/\[+\]//' | xargs rm
fdupes -NdS extra | grep '[+]' | sed 's/\[+\]//' | xargs rm

# Remove the old set of supported languages
rm LINGUAS extra/LINGUAS

# Generate a new set with all the po files that were not excluded
ls --recursive *.po | sed 's/.po//g' | sort -udo LINGUAS

#Remove trailling new Line
perl -pi -e 'chomp if eof' LINGUAS

# Copy the file from the po directory to po/extra
cp LINGUAS extra/
```


po/extra/LINGUAS
```diff
@@ -1,196 +1,60 @@
-aa
-ab
-ae
-af
 ak
```
ryonakano: It seems `po/extra/ak.po` is removed but its language code still stays here. See https://travis-ci.com/elementary/mail/builds/114816473#L3951:

```
msgfmt: error while opening "ak.po" for reading: No such file or directory
```

cjfloss: TL;DR 
Running `ninja extra-update-po` and `ninja meson-pantheon-mail-update-po` fixes it

----

The problem is that extra did not have translations but the main ones have, so extra/*.po got removed.
```
po/extra/ak.po
po/extra/bs.po
po/extra/en_CA.po
po/extra/eo.po
po/extra/eu.po
po/extra/fa.po
po/extra/gl.po
po/extra/hi.po
po/extra/kk.po
po/extra/km.po
po/extra/my.po
po/extra/sr@latin.po
po/extra/te.po
po/extra/th.po
po/extra/ur.po
```

The script removes `po/LINGUAS` & `po/extra/LINGUAS`, afterwards it gets all the languages that have some translation and create a new LINGUAS file and places a copy in the `po` directory and in `po/extra`.

My thinking behind this is: If the main one have translation it is probable that someone wants to translate for that language, so it uses the same LINGUAS file for both main and extra.

With this PR I wanted to do just the removal, but if you want I can update the po files as well.

tintou:
Everything is now handled by Weblate so I'm going to close this

PR #392, opened 2019-06-12
Wait to parse mailto until ComposerWidget

danrabbit:
Doing it this way we simplify the number of arguments we're passing by quite a bit and we keep ComposerWidget to be a more strict GObject style

Also add BCC support

### DIFF
--- a/src/Composer/ComposerWidget.vala
+++ b/src/Composer/ComposerWidget.vala
@@ -46,7 +46,6 @@ public class Mail.ComposerWidget : Gtk.Grid {
     private Gtk.Entry cc_val;
     private Gtk.Entry bcc_val;
     private Gtk.Revealer cc_revealer;
-    private Gtk.ToggleButton bcc_button;
     private Gtk.ToggleButton cc_button;
     private Granite.Widgets.OverlayBar message_url_overlay;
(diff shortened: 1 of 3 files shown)


src/Composer/ComposerWidget.vala
```diff
@@ -358,6 +352,36 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+            var result = new Gee.HashMap<string, string> ();
+            var params = mailto_query.split ("&");
+
+            foreach (unowned string param in params) {
+                var terms = param.split ("=");
+                if (terms.length != 2) {
+                    critical ("Invalid mailto URL");
+                }
+
+                result[terms[0]] = Soup.URI.decode (terms[1]);
```
vertexian: This should be in an `else` clause, I think. I think it might even be a crash if `terms.length < 2`?

danrabbit: Nice catch!

vertexian:
Looks good now. I'm ambivalent about whether the parsing logic goes in the widget or the application controller.

danrabbit:
Discussed with Corentin in Slack that this is also where Evolution does the parsing so it seems to make sense :)

Issue #261, opened 2018-10-17
Print a message


danrabbit:
We lost the dropdown that had the print function in it

vertexian:
Should finish with this today. I'm also adding back the per message options menu, with reply, reply all, forward, and mark read/unread commands, in addition to printing.

marbetschar:
@danrabbit is this still relevant? I can see the "More" menu with a "Print..." option in the message header?

cblock:
I can confirm that in v6.4.0 the dropdown exists. So I guess, this issue could be closed

danrabbit:
Yup closing :)

PR #399, opened 2019-06-24
#261: add back per-message actions menu, including print

vertexian:
Adds an actions pop-up menu to each message, allowing:
1. Reply
2. Reply All
3. Forward
4. Print

Unlike with the buttons in the window header, these actions apply only to a single message.

There are some limitations:
1. I wasn't able to add Mark as Read/Unread because of a problem with action enabled state. (There's a related bug in the existing button in the window header).
2. The reply/forward menu items don't disable when a composer is already open (see comment below). However, they won't open a second composer either.
3. We don't scroll the composer into view, because there's a hack for this that's awkward to access from `MessageListItem`. 
4. We don't show header fields in printed output. I'm trying to follow Geary's approach for this (injecting HTML and CSS), but left it out of this PR.


src/MessageList/MessageListBox.vala
```diff
@@ -97,30 +97,33 @@ public class Mail.MessageListBox : Gtk.ListBox {
...
+        }
+
+        if (message_item == null) {
+            message_item = (MessageListItem) last_child;
+        }
+
         string content_to_quote = "";
         Camel.MimeMessage? mime_message = null;
         Camel.MessageInfo? message_info = null;
-        if (last_child is MessageListItem) {
```
vertexian: Not possible for `last_child` to be neither `InlineComposer` nor `MessageListItem`.

src/MessageList/MessageListBox.vala
```diff
@@ -97,30 +97,33 @@ public class Mail.MessageListBox : Gtk.ListBox {
...
+        var composer = new InlineComposer (type, message_info, mime_message, content_to_quote);
+        composer.discarded.connect (() => {
+            can_reply = true;
             can_move_thread = true;
-        }
+            remove (composer);
+            composer.destroy ();
+        });
+        add (composer);
+        can_reply = false;
```
vertexian: Should I bind to this from the menu in `MessageListItem`, in order to set the enabled state on the action menu items? That feels awkward.

src/MessageList/MessageListItem.vala
```diff
@@ -35,7 +35,7 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
     private string message_content;
     private bool message_is_html = false;
     private bool message_loaded = false;
-
```
vertexian: Fixes lint issue.

src/MessageList/MessageListItem.vala
```diff
@@ -290,6 +291,85 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+            print_operation.set_print_settings (settings);
+            print_operation.run_dialog ((Gtk.ApplicationWindow) get_toplevel ());
+        } catch (Error e) {
+            critical ("Print operation failed. %s".printf (e.message));
+        }
+    }
+
+    /* Replace any runs of whitespace, non-printing characters or slashes with a
+       single space and remove and leading or trailing spaces. */
+    private string safe_filename (string str, ssize_t max_length) throws Error {
```
vertexian: Adapted from Geary, with some tweaks. Could also be used if we add a feature for saving emails as .eml files.

src/MessageList/MessageListItem.vala
```diff
@@ -290,6 +291,85 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+            menu_button.set_active (true);
+            return Gdk.EVENT_STOP;
+        });
+
+        menu_button.show ();
+        return menu_button;
+    }
+
+    private void add_inline_composer (ComposerWidget.Type composer_type) {
+        var message_list_box = (MessageListBox) get_parent();
```
vertexian: It looks like Mail 1.0.8 uses an approach where `MessageListItem` has a signal for reply, which chains to a signal in `MessageListBox`, which ultimately has a handler in `MainWindow`. That might be cleaner, and solve the problem of how to implement the hack where we scroll the inline composer into view.

jlnr:
Sorry to force push onto your branch @vertexian, but this should clear up the broken merge commits that didn't build.

src/MessageList/MessageListItem.vala
```diff
@@ -290,6 +291,85 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+        message_list_box.add_inline_composer (composer_type, this);
+    }
+
+    private void on_print () {
+        try {
+            Gtk.PrintSettings settings = new Gtk.PrintSettings ();
+            string filename = _("Email Message");
+
+            string subject = message_info.subject;
```
jlnr: I think this string can be `unowned`

src/MessageList/MessageListItem.vala
```diff
@@ -290,6 +291,85 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+
+    private void add_inline_composer (ComposerWidget.Type composer_type) {
+        var message_list_box = (MessageListBox) get_parent();
+        message_list_box.add_inline_composer (composer_type, this);
+    }
+
+    private void on_print () {
+        try {
+            Gtk.PrintSettings settings = new Gtk.PrintSettings ();
+            string filename = _("Email Message");
```
jlnr: Would be good to have a comment for translators that this is a default filename.

src/MessageList/MessageListItem.vala
```diff
@@ -290,6 +291,85 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+        } catch (Error e) {
+            critical ("Print operation failed. %s".printf (e.message));
+        }
+    }
+
+    /* Replace any runs of whitespace, non-printing characters or slashes with a
+       single space and remove and leading or trailing spaces. */
+    private string safe_filename (string str, ssize_t max_length) throws Error {
+        var s = new Regex ("[[:space:][:cntrl:]/]+").replace (str, -1, 0, " ").strip ();
+        if (s.length < max_length) {
```
jlnr: Not that any user will notice, but on a code level I think <= would be more intuitive (max_length being the longest length that is allowed).

danrabbit:
hey @vertexian thanks for your work on this. I resolved conflicts with master and made the linter happy in another branch and we got this merged in :)

PR #412, opened 2019-08-15
Use correct sidebar icons

### LINKED ISSUE #262: Folder icon detection [$50]
It looks like we don't show the proper icons for several folders. I'm not sure if this just needs to be hooked up from what EDS gives us or we need some kind of heuristic. It looks like the only one that Evolution picks up that we don't currently in "Drafts"

screenshot from 2018-10-17 14 15 30 2x

<bountysource-plugin>

---
There is a **[$50 open bounty](https://www.bountysource.com/issues/64895548-folder-icon-detection?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oct2pus:
This should fix #262, although a better solution could be available eventually.

This solution is very inelegant, to the point I feel I need to explain why I did it this way. Icons are handled and assigned by FolderSourceItem.update_infos(). update_infos() follows the [advice of the libcamel documentation](https://developer.gnome.org/camel/stable/camel-camel-enums.html#CamelFolderInfoFlags) and compares the binary AND result of both folderinfo.flags and Camel.FOLDER_TYPE_MASK to the enum Camel.FolderInfoFlags.TYPE_X values. The problem is this value isn't returning the expected number for some of the values, this results some icons using the "folder-documents" icon.

I checked to see how evolution handles it and Evolution just checks the names for 5 of the values, using a boolean for 2 of them to track if its true or not, and the 3 others they supply a modified value (if i'm understanding my boolean math correctly). Evolution then sends all other results to a different function with a similar switch statement to what Mail was already doing.

The relevant evolution code is in [em-folder-tree-model.c](https://gitlab.gnome.org/GNOME/evolution/blob/2fa97de820634ac002784165bb122bc32e651235/src/mail/em-folder-tree-model.c#L691) and [em-folder-utils.c](https://gitlab.gnome.org/GNOME/evolution/blob/2fa97de820634ac002784165bb122bc32e651235/src/mail/em-folder-utils.c#L596). Hopefully I didn't miss anything in Evolution. I suspect that something is up in libcamel, but I did keep the original comparisons as well. This would provide for a slightly more robust behavior if libcamel ever gets fixed, and a basic "fallback" behavior if its not.

image

edit: fixed a typo.

### DIFF
--- a/src/FoldersView/FolderSourceItem.vala
+++ b/src/FoldersView/FolderSourceItem.vala
@@ -62,9 +62,8 @@ public class Mail.FolderSourceItem : Granite.Widgets.SourceList.ExpandableItem {
         }
 
         
-        // Camel.FolderInfo.Flags does not return expected values for multiple types of mail folders
-        // after being processed by bitwise AND the Camel.FOLDER_TYPE_MASK. Original comparisons are
-        // preserved if libcamel is ever updated to address this issue.
+        // (Camel.FolderInfo.Flags & Camel.FOLDER_TYPE_MASK) does not return expected values for multiple types of folders.
+        // Original comparisons are preserved if/when this bug is fixed in libcamel.
         var masked_flag = folderinfo.flags & Camel.FOLDER_TYPE_MASK;
         var lowercase_full_name = full_name.down ();
 
--- a/src/FoldersView/FolderSourceItem.vala
+++ b/src/FoldersView/FolderSourceItem.vala
@@ -67,7 +67,7 @@ public class Mail.FolderSourceItem : Granite.Widgets.SourceList.ExpandableItem {
         // preserved if libcamel is ever updated to address this issue.
         var masked_flag = folderinfo.flags & Camel.FOLDER_TYPE_MASK;
         var lowercase_full_name = full_name.down ();
-        
+
         if (masked_flag == Camel.FolderInfoFlags.TYPE_INBOX ||
         lowercase_full_name.contains ("inbox")) {
             icon = new ThemedIcon ("mail-inbox");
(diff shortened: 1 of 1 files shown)


src/FoldersView/FolderSourceItem.vala
```diff
@@ -61,41 +61,46 @@ public class Mail.FolderSourceItem : Granite.Widgets.SourceList.ExpandableItem {
...
-                can_modify = true;
-                break;
+        
+        // (Camel.FolderInfo.Flags & Camel.FOLDER_TYPE_MASK) does not return expected values for multiple types of folders.
+        // Original comparisons are preserved if/when this bug is fixed in libcamel.
+        var masked_flag = folderinfo.flags & Camel.FOLDER_TYPE_MASK;
+        var lowercase_full_name = full_name.down ();
+
+        if (masked_flag == Camel.FolderInfoFlags.TYPE_INBOX ||
+        lowercase_full_name.contains ("inbox")) {
```
jlnr: Formatting nitpick: I'm not sure if there is an elementary guideline for this, but I think these continued `if` conditions should either be indented with 4 or 8 spaces, not 0 relative to the `if` statement.

oct2pus: I just found the elementary Code guidelines and i don't believe it does. This is not really a normal case to begin with so I can see why this might be missed.

oct2pus:
I can't believe I forgot about languages that aren't english, oops.

 i'll take some time today or tomorrow to work on this, hopefully it shouldn't be too complex of a fix :sweat_smile:

tintou:
Closing stale merge request, feel free to reopen it once it is ready for review :tada:

Issue #240, opened 2018-09-18
Tapping Esc when drafting a mail doesn't save the draft


cassidyjames:
As far as I can tell, hitting the close button on the window saves the draft, but hitting the escape key does not. I just accidentally lost a big long draft. :cry:

WattanaGaming:
If you ware talking about the 0.4 Loki, then it doesnt save the draft *at all*

danrabbit:
As far as I can tell, in the mail rewrite drafts are never saved

vertexian:
So what is the design we want for drafts? It sounds like we want auto-save on close, but should it be auto-saving regularly all the time? Or do we want a Save action in the UI?

cassidyjames:
@vertexian following the elementary HIG would imply autosaving _at least_ on close, if not at regular intervals/when things change (like Code).

getsnoopy:
This seems to be current behaviour, at least on 5.1 Hera.

ramonziai:
I can confirm this behavior on 5.1 Hera. Just lost a rather complicated draft because I accidentally hit ESC and now I can't seem to find it anywhere. [rantmode] Honestly, I don't understand how one can come up with this behavior in the first place, seriously considering switching back to Thunderbird or Evolution right now...[/rantmode]

Bounty upped to $30. Is there anything else one can do to speed up fixing this? It's a deal-breaker for me, and I otherwise rather like this application.

ramonziai:
Just had a look at the code (I've never written a line of Vala in my life, but it seems similar enough to Java). There doesn't seem to be a method for saving drafts at all, unless I missed something. If so, this seems to be a bigger change. How about meanwhile introducing a little confirmation dialog whenever people hit ESC or the "Discard Draft" button? Code also asks for confirmation if you attempt to close a new unnamed file, so that shouldn't conflict with the HIG.

Also, how about hiding "Drafts" from the folder view until there's an established way of actually putting something there?

danrabbit:
Draft saving when closing the composer is now implemented, however we don't currently close the compose window when pressing the escape key. So I'm going to leave this issue open for that, but bumping from milestones since the critical issue of saving drafts has been fixed

PR #458, opened 2020-02-19
Added discard draft warning dialog

codemessiah:
I have attempted to solve issue #240 by showing a warning dialog when a user presses the Escape key or clicks the tiny trashcan icon at the bottom left of the composer dialog. I believe the behaviour of Escape and clicking the trashcan icons are semi-ambiguous and would result in most users learning their true functionality the hard way. This dialog clarifies the intentions of the actions.

Screenshot from 2020-02-18 22-30-41


src/Composer/ComposerWidget.vala
```diff
@@ -490,7 +490,26 @@ public class Mail.ComposerWidget : Gtk.Grid {
     }
 
     private void on_discard () {
-        discarded ();
+        var discard_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("Delete This Draft?"), // choice of 'Delete' over 'Discard' for clarity
+            _("Are you sure you want to delete this draft permanently?"),
+            "dialog-error",
```
danrabbit: Since this isn't an error message `dialog-error` is probably not appropriate here
```suggestion
            "dialog-warning",
```

src/Composer/ComposerWidget.vala
```diff
@@ -490,7 +490,26 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+            Gtk.ButtonsType.NONE
+        );
+        discard_dialog.transient_for = get_toplevel () as Gtk.Window;
+
+        discard_dialog.add_button (_("No"), Gtk.ResponseType.CANCEL);
```
danrabbit: Action buttons should never be "Yes", "No", or "OK". They should always describe an explicit action. In this case, it would be much better to use "Cancel"
```suggestion
        discard_dialog.add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
```

src/Composer/ComposerWidget.vala
```diff
@@ -490,7 +490,26 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+
+        var discard_anyway = discard_dialog.add_button (_("Yes"), Gtk.ResponseType.ACCEPT);
```
danrabbit: Same as above, this should be an explicit action name
```suggestion
        var discard_anyway = discard_dialog.add_button (_("Delete Draft"), Gtk.ResponseType.ACCEPT);
```

src/Composer/ComposerWidget.vala
```diff
@@ -490,7 +490,26 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+        discard_anyway.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+
+        if (discard_dialog.run () == Gtk.ResponseType.CANCEL) { // User does NOT want to discard draft
+            discard_dialog.destroy ();
+            return;
+        }
+        discard_dialog.destroy ();
+
+        discarded (); // Raise signal
```
danrabbit: Seems like you're repeating yourself :) Maybe instead this should be:

```
        if (discard_dialog.run () == Gtk.ResponseType.ACCEPT) {
            discarded ();
        }

        discard_dialog.destroy ();
 ```

File: src/Composer/ComposerWidget.vala [hunk 5]
```diff
@@ -490,7 +490,26 @@ public class Mail.ComposerWidget : Gtk.Grid {
     }
 
     private void on_discard () {
-        discarded ();
+        var discard_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("Delete This Draft?"), // choice of 'Delete' over 'Discard' for clarity
+            _("Are you sure you want to delete this draft permanently?"),
```
danrabbit: Instead of repeating yourself here, this should offer some kind of further explanation. You might want to try answering additional questions like "Will my draft be deleted from the server too or just this computer?" and "Will I be able to undo this decision?"

src/Composer/ComposerWidget.vala
```diff
@@ -490,7 +490,26 @@ public class Mail.ComposerWidget : Gtk.Grid {
     }
 
     private void on_discard () {
-        discarded ();
+        var discard_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("Delete This Draft?"), // choice of 'Delete' over 'Discard' for clarity
```
danrabbit: Dialog titles should be sentence case
```suggestion
            _("Permanently delete this draft?"),
```

codemessiah:
Hey, thanks for the code review! I’ll make your suggested changes when I get home from work tonight.

src/Composer/ComposerWidget.vala
codemessiah: I'm not able to answer that first question because I don't actually know much about the Mail app (because I'm never able to login with my Gmail). I will do my best beyond that, though

PR #472, opened 2020-04-18
webkit-extension: Do not use the deprecated DOM API

tintou:
Remove the use of deprecated API and use Javascriptcore instead API instead

### DIFF
--- a/webkit-extension/MailPage.vala
+++ b/webkit-extension/MailPage.vala
@@ -34,69 +34,73 @@ public class Mail.Page : Object {
     }
 
     private bool on_page_user_message_received (WebKit.WebPage page, WebKit.UserMessage message) {
+        var js_context = page.get_main_frame ().get_js_context ();
         switch (message.name) {
             case "set-body-html":
-                unowned string body = message.parameters.get_string ();
-                try {
-                    page.get_dom_document ().get_document_element ().query_selector ("#message-body").set_inner_html (body);
-                    return true;
-                } catch (Error e) {
-                    warning ("Unable to set message body content: %s", e.message);
-                }
-
-                return false;
+                unowned string body_html = message.parameters.get_string ();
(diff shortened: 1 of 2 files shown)


webkit-extension/MailPage.vala
```diff
@@ -34,69 +34,73 @@ public class Mail.Page : Object {
...
-                try {
-                    page.get_dom_document ().get_document_element ().query_selector ("#message-body").set_inner_html (body);
-                    return true;
-                } catch (Error e) {
-                    warning ("Unable to set message body content: %s", e.message);
-                }
-
-                return false;
+                unowned string body_html = message.parameters.get_string ();
+                var body = js_context.evaluate ("document.querySelector(\"#message-body\")", -1);
```
davidmhewitt, line 41: Single quotes and double quotes are interchangeable in JS. I'd be tempted to use single quotes thoughout this so we don't have to see escape backslashes in the embedded JS.

webkit-extension/MailPage.vala
```diff
@@ -34,69 +34,73 @@ public class Mail.Page : Object {
...
-                try {
-                    var body_html = page.get_dom_document ().get_document_element ().query_selector ("body").get_inner_html ();
-                    message.send_reply (new WebKit.UserMessage ("get-body-html", new Variant.take_string ((owned) body_html)));
-                    return true;
-                } catch (Error e) {
-                    warning ("Unable to get message body content: %s", e.message);
-                }
-
-                return false;
+                JSC.Value val = js_context.evaluate ("document.querySelector(\"body\").innerHTML;", -1);
```
davidmhewitt, line 45: Same here

webkit-extension/MailPage.vala
tintou: true

PR #475, opened 2020-04-24
Prefer dark

danrabbit:
- [ ] What to do (anything?) about mail contents?

It turns out macOS supports the WebKit media query in Mail, and also will automatically attempt to invert "simple" mail unless it's told otherwise: https://webkit.org/blog/8840/dark-mode-support-in-webkit/

My inclination is to do nothing for now tbh. It would be nice if HTML mail senders include dark styles, but at least the UI and the composer works


danrabbit:
Needs to be updated for granite.settings

danrabbit:
The composer also needs a dark mode style

Screenshot from 2021-01-28 15 14 11@2x

cassidyjames (approved):
lfg

PR #484, opened 2020-05-21
Search should be insensitive with no accounts added #319 

### LINKED ISSUE #319: Search should be insensitive with no accounts added
What it says on the tin. It's currently sensitive all the time, whether or not any accounts are added.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65447794-search-should-be-insensitive-with-no-accounts-added?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github).
</bountysource-plugin>

JO0st:
fixes #319

### DIFF
--- a/src/Backend/Session.vala
+++ b/src/Backend/Session.vala
@@ -41,6 +41,8 @@ public class Mail.Backend.Session : Camel.Session {
 
     public signal void account_added (Mail.Backend.Account account);
 
+    public signal void account_removed (Mail.Backend.Account account);
+
     private Session () {
         Object (user_data_dir: Path.build_filename (E.get_user_data_dir (), "mail"), user_cache_dir: Path.build_filename (E.get_user_cache_dir (), "mail"));
     }
@@ -250,6 +252,12 @@ public class Mail.Backend.Session : Camel.Session {
         }
 
         return service;
+    }
+
+    public override void remove_service (Camel.Service service) {
+        var account = new Account (service);
+        base.remove_service (service);
+        account_removed (account);
     }
 
     public Gee.LinkedList<Backend.Account> get_accounts () {
--- a/src/Backend/Session.vala
+++ b/src/Backend/Session.vala
@@ -41,7 +41,7 @@ public class Mail.Backend.Session : Camel.Session {
 
     public signal void account_added (Mail.Backend.Account account);
 
-    public signal void account_removed (Mail.Backend.Account account);
+    public signal void account_removed ();
 
     private Session () {
         Object (user_data_dir: Path.build_filename (E.get_user_data_dir (), "mail"), user_cache_dir: Path.build_filename (E.get_user_cache_dir (), "mail"));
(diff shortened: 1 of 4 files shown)


davidmhewitt:
Hey, thanks for working on this!

I believe it should be possible to use [bind_property](https://valadoc.org/gobject-2.0/GLib.Object.bind_property.html) to simplify this.

You'd have your property like:
`public bool can_search { get; private set; }`

Then set up `bind_property` something like:
`bind_property ("can-search", search_entry, "sensitive", BindingFlags.SYNC_CREATE)`

This would then cause the `sensitive` property on the search box to be updated every time the `can_search` property was updated.

Do you want to try that out and see if you can get it to work? I'm not currently at my dev PC so I couldn't test it out.

JO0st:
thanks for the hint. I tried it and it works, it is included in the commit to make the search insensitive when the last account is removed.

src/Backend/Session.vala
```diff
@@ -252,6 +254,12 @@ public class Mail.Backend.Session : Camel.Session {
         return service;
     }
 
+    public override void remove_service (Camel.Service service) {
+        var account = new Account (service);
```
davidmhewitt, line 258: I'm not sure we need to construct an `Account` here and pass it as a parameter to the signal as it's never used.

src/HeaderBar.vala
```diff
@@ -20,7 +20,8 @@
 
 public class Mail.HeaderBar : Gtk.HeaderBar {
     public bool can_mark { get; set; }
-    public Gtk.SearchEntry search_entry { get; construct; }
+    public bool can_search { get; set; }
+    public Gtk.SearchEntry search_entry { get; set; }
```
davidmhewitt, line 24: Another thing I've just spotted, it shouldn't be necessary to change the `construct` to `set` here.

src/Backend/Session.vala
JO0st: The reason I allready put it in here is to make sure the signal can be used in other places. For example the account list could use this to remove an account when necessary, but to do that it needs to know which account to remove. Should i be proactive and put it in there, or leave it out until it is necessary for another usecase?

davidmhewitt: I'd say leave it out until it is necessary. Since you're constructing a new object here, it's not the same `Account` as one constructed elsewhere, so it may not be that useful for comparisons to already existing `Account` objects.

JO0st: OK, I have removed the parameter from the signal. Thanks for the thorough review

danrabbit:
Hey thanks for your work on this feature. I have resolved conflicts with this branch in https://github.com/elementary/mail/pull/502 and merged it!

PR #500, opened 2020-07-16
ConversationListItem: Use accent color for unread badge

danrabbit:
Does this need to be blue or do we actually want it to be accent color?


cassidyjames:
Ooh, I like it being accent-colored, I think!

fabiozaramella (approved):

PR #506, opened 2020-08-20
Make Compose button insensitive with no accounts

### LINKED ISSUE #320: Compose button should be insensitive with no accounts
What it says on the tin. It's currently sensitive all the time, whether or not any accounts are added.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65447793-compose-button-should-be-insensitive-with-no-accounts?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dar5hak:
Fixes #320


dar5hak:
Okay, this should be a simple fix in theory, but I can't for the life of me figure out why my code doesn't work in its current state.

**What I did:** emulate `can_search` behavior in a new variable called `can_compose`. I expected that to make the Compose button insensitive whenever the search entry is insensitive.

**What really happens:** Compose button is sensitive at launch (with no accounts). It only becomes insensitive right when I quit the application. I tested this using `notify` and `printf`.

**What options I tried:** I tried binding the Compose button directly to `can_search`, expecting it to exactly follow the search entry behaviour. And it still stayed sensitive.

Can someone help me fix this?

dar5hak:
Don't know if it's relevant but I'm using Granite v5.5.0 on latest Hera (not Odin Early Access).

dar5hak:
Update: doesn't work on Odin daily either. There's surely something wrong with my code.

fs3o5:
I just tried to see what's happening here and it seems setting an `action_name` conflicts with setting `sensitive = false`.
I don't really know why that's the case though.

Creating the compose button like this
```
can_compose = false;
var compose_button = new Gtk.Button.from_icon_name ("mail-message-new", Gtk.IconSize.LARGE_TOOLBAR) {
    sensitive = false,
    halign = Gtk.Align.START
};
```

and syncing the state like this
```
notify["can-compose"].connect (() => {
    compose_button.action_name = can_compose ? MainWindow.ACTION_PREFIX + MainWindow.ACTION_COMPOSE_MESSAGE : null;
    compose_button.sensitive = can_compose;
});
```
should work.

dar5hak:
@fs3o5 Whoa, thanks. That worked. Indeed, it's strange for the sensitivity to conflict with `action_name`.

ryonakano (changes requested):
Hi, thank you for your contribution! I have two comments:

- Now the compose_button gets insensitive depends on the existence of accounts, but the shortcut action `Ctrl+N` still works even if you have no accounts
- I get the following error message on terminal:
```
(io.elementary.mail:16818): Gtk-CRITICAL **: 09:33:17.182: gtk_application_get_accels_for_action: assertion 'detailed_action_name != NULL' failed
```

@dar5hak I'll create a PR that contains suggested changes so would you merge it in [your folked repo](https://github.com/dar5hak/mail.git)?

dar5hak:
Thanks for finding this out, Ryo-san. Merged it.

ryonakano (approved):
Works for me.

src/MainWindow.vala
```diff
@@ -210,12 +211,14 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
         unowned Mail.Backend.Session session = Mail.Backend.Session.get_default ();
         session.account_added.connect (() => {
             placeholder_stack.visible_child = paned_end;
+            headerbar.can_compose = true;
```
tintou, line 214: I don't think that the rest is necessary, the button state will be synced with the action sate
```suggestion
            get_action (ACTION_COMPOSE_MESSAGE).set_enabled (false);
```

src/MainWindow.vala
```diff
@@ -210,12 +211,14 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
...
             headerbar.can_search = true;
         });
 
         session.account_removed.connect (() => {
             var accounts_left = session.get_accounts ();
             if (accounts_left.size == 0) {
+                headerbar.can_compose = false;
```
tintou, line 221:
```suggestion
            get_action (ACTION_COMPOSE_MESSAGE).set_enabled (false);
```

dar5hak:
Okay, hold on. The button stays sensitive or insensitive like it should. But Ctrl+N doesn't work even when there is an account. This happens both with and without @tintou's suggested changes.

dar5hak:
Bummer. Seems like a problem with my keyboard. Ctrl+N combo doesn't work in Firefox and VSCodium either.

Can someone test Ctrl+N accel for this? If it works, this is ready to be merged.

ryonakano:
I tested this yesterday but Ctrl+N works as expected for me.

tintou:
In fact, you don't even need all that property bindings.
Here is the only diff required:
```diff
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -88,6 +88,7 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
 
     construct {
         add_action_entries (ACTION_ENTRIES, this);
+        get_action (ACTION_COMPOSE_MESSAGE).set_enabled (false);
 
         foreach (var action in action_accelerators.get_keys ()) {
             ((Gtk.Application) GLib.Application.get_default ()).set_accels_for_action (
@@ -210,12 +211,14 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
         unowned Mail.Backend.Session session = Mail.Backend.Session.get_default ();
         session.account_added.connect (() => {
             placeholder_stack.visible_child = paned_end;
+            get_action (ACTION_COMPOSE_MESSAGE).set_enabled (true);
             headerbar.can_search = true;
         });
 
         session.account_removed.connect (() => {
             var accounts_left = session.get_accounts ();
             if (accounts_left.size == 0) {
+                get_action (ACTION_COMPOSE_MESSAGE).set_enabled (false);
                 headerbar.can_search = false;
             }
         });
```

dar5hak:
Mother of Merlin, I did not know actions would simply sync with the button sensitivity. That's super convenient. 😮️

tintou (approved):

tintou:
Thanks for this :slightly_smiling_face:

PR #511, opened 2020-10-04
MessageListItem: use Hdy.Avatar

### LINKED ISSUE #510: Use Hdy.Avatar

danrabbit:
Fixes #510

### DIFF
--- a/src/MessageList/MessageListItem.vala
+++ b/src/MessageList/MessageListItem.vala
@@ -68,10 +68,23 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         loading_cancellable = new GLib.Cancellable ();
 
         style_context = get_style_context ();
-        style_context.add_class ("card");
+        style_context.add_class (Granite.STYLE_CLASS_CARD);
 
-        var avatar = new Granite.Widgets.Avatar.with_default_icon (48);
-        avatar.valign = Gtk.Align.START;
+        string parsed_address;
+        string parsed_name;
+
+        var camel_address = new Camel.InternetAddress ();
+        camel_address.unformat (message_info.from);
+        camel_address.ref ();
+        camel_address.get (0, out parsed_name, out parsed_address);
+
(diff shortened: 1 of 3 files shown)


src/MessageList/MessageListItem.vala
```diff
@@ -68,10 +68,23 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         loading_cancellable = new GLib.Cancellable ();
 
         style_context = get_style_context ();
-        style_context.add_class ("card");
+        style_context.add_class (Granite.STYLE_CLASS_CARD);
 
-        var avatar = new Granite.Widgets.Avatar.with_default_icon (48);
-        avatar.valign = Gtk.Align.START;
+        string parsed_address;
```
tintou, line 73:
```suggestion
        string? parsed_address;
```

src/MessageList/MessageListItem.vala
```diff
@@ -68,10 +68,23 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+        string parsed_name;
```
tintou, line 74:
```suggestion
        string? parsed_name;
```

src/MessageList/MessageListItem.vala
```diff
@@ -68,10 +68,23 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+
+        var camel_address = new Camel.InternetAddress ();
+        camel_address.unformat (message_info.from);
+        camel_address.ref ();
```
tintou, line 78: You're leaking a reference here
```suggestion
```

tintou (changes requested):

src/MessageList/MessageListItem.vala
danrabbit: Without the ref I get a crash from double free

tintou (approved):

PR #526, opened 2020-12-21
Automatically mark messages as read

### LINKED ISSUE #278: Messages aren't automatically marked as read
There was some kind of heuristic before that would mark a conversation as being read automatically upon viewing. It doesn't look like conversations are ever marked as read in the rewrite yet.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/64937318-messages-aren-t-automatically-marked-as-read?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #278 

Does 5 seconds seem like a reasonable value?

### DIFF
--- a/src/ConversationList/ConversationListBox.vala
+++ b/src/ConversationList/ConversationListBox.vala
@@ -24,6 +24,8 @@ public class Mail.ConversationListBox : VirtualizingListBox {
     public signal void conversation_selected (Camel.FolderThreadNode? node);
     public signal void conversation_focused (Camel.FolderThreadNode? node);
 
+    private const int MARK_READ_TIMEOUT_SECONDS = 5;
+
     public Backend.Account current_account { get; private set; }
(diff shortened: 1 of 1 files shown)


danrabbit:
That seems more than reasonable to me. It looks like iOS mail does like 2 seconds

danrabbit (approved):
It looks like Geary had some more complicated behavior like marking an individual message as read based where it was in the view and things like this, but a simple timeout solves the problem IMO. We can always revisit with a more complicated behavior if it turns out to be a real problem

PR #537, opened 2021-01-05
RDNN gettext name

ryonakano:
This seems to cause the UI not shown as translated


danrabbit (changes requested):
Probably also need to `git mv` the current POT file

danrabbit (approved):

PR #539, opened 2021-01-05
Fix "Message Deleted" not shown as translated

ryonakano:
## Before
Screenshot from 2021-01-05 14-35-33

## After
Screenshot from 2021-01-05 14-38-34


danrabbit (approved):

tintou:
that's not good, that means that we're missing a step for the translation at some place

PR #541, opened 2021-01-05
ConversationListItem: Add flat class to badge label

tintou:
As per https://github.com/elementary/stylesheet/issues/876


src/ConversationList/ConversationListItem.vala
```diff
@@ -53,6 +53,7 @@ public class Mail.ConversationListItem : VirtualizingListBoxRow {
         weak Gtk.StyleContext messages_style = messages.get_style_context ();
         messages_style.add_class (Granite.STYLE_CLASS_BADGE);
         messages_style.add_class (Gtk.STYLE_CLASS_SIDEBAR);
```
danrabbit, line 55: I guess we can remove this, since it's probably a workaround for what flat class is intended to do

danrabbit (approved):

PR #553, opened 2021-03-14
Add window title

JoseExposito:
At the moment, the main window doesn't have a title. It is not a big deal since the title is not displayed anywhere in the UI.
However, once https://github.com/elementary/gala/issues/798 gets fixed, Gala will display "io.elementary.mail" as window title.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -87,6 +87,8 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
     }
 
     construct {
+        title = _("Mail");
+
         add_action_entries (ACTION_ENTRIES, this);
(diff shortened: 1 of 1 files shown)


cassidyjames:
Window titles are also used in the Plank dock menu in the case of multiple windows, and generally should just be set no matter what in case they are used/displayed elsewhere. 👍

cassidyjames (changes requested):
Reviewing on mobile, so apologies if I am missing something. But I think best practice is to set this in the object construction around line 69.

JoseExposito:
> Window titles are also used in the Plank dock menu in the case of multiple windows

Good point, @cassidyjames I didn't know they're displayed there. I made the changes you requested :smile:

cassidyjames (approved):

Issue #449, opened 2020-01-12
Grouped Folders


marbetschar:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature

If we do have multiple email accounts in elementary Mail it would be convenient to add grouped folders - at least for well known folder names. That is: Inbox, Archive, Drafts, Sent. I would expect these folders to be shown at the very top, containing the messages from all configured email accounts. This way we can get one single Inbox folder, showing an overview for all received messages, one single Archive folder showing all archived messages, ... etc.

**Existing work**
Basically everything is there. Would just need an additional check if there are multiple accounts, and if so show another, predefined group of folders which are linked to the corresponding folders in all accounts.

PR #545, opened 2021-02-18
Folder icon detection

marbetschar:
Fixes #262.

Implemented similar to [how it is done in Evolution](https://gitlab.gnome.org/GNOME/evolution/-/blob/2fa97de820634ac002784165bb122bc32e651235/src/mail/em-folder-tree-model.c#L692-771).

However, this PR does not cover the following:

- Icon detection for Templates folder (because there is no `Camel.FolderInfoFlags.TYPE_TEMPLATE`)
- Icon detection for local folders (seemed like a legacy thing, so I did not spent any time on this - [if someone needs this, here is how its done in Evolution](https://gitlab.gnome.org/GNOME/evolution/-/blob/2fa97de820634ac002784165bb122bc32e651235/src/mail/em-folder-tree-model.c#L726-738))

PR #564, opened 2021-04-29
Grouped Folders: Inbox

marbetschar:
This PR eventually fixes https://github.com/elementary/mail/issues/449.

Right now it only adds a unified Inbox, however we may want to iterate on this a bit more to expand support to other well known folders if possible (such as Drafts, Outbox, Sent, Archive, Junk and Trash).

Please note the current implementation uses a hardcoded folder named "INBOX". While this seems to work for this particular folder, this does not seem to be a reliable approach for other well known folders. There is already a [open PR](https://github.com/elementary/mail/pull/545) which provides an improved folder detection mechanism - so that's probably a good starting point.

Screenshot from 2021-05-04 07-56-03


danrabbit:
I can confirm that this works as expected! Can we add this to a header like "All Mailboxes" so it's a bit more clear what this is?

marbetschar:
@danrabbit added a "All Mailboxes" section header.

hanaral:
Would it be possible to add an 'all sent' button as well? iOS Mail has the ability to add/remove quite a few different unified section and I think it would be a good starting point to add the main 2

danrabbit:
@hanaral We usually prefer to have many smaller PRs than to try to do everything at once in a single branch. This makes code review faster and easier and helps with finding issues down the road if they occur

marbetschar:
@danrabbit / @hanaral I digged into this a bit more and added support for `INBOX`, `Drafts`, `Sent` and `Archive` folders. Please note, showing a badge with unread messages is still not supported yet.

danrabbit (changes requested):
Inbox works here, but I'm getting a number of errors for the other mailboxes. I think it would be best to go back to just inbox in this branch and then do follow up branches for other other mailbox types

on launch I'm getting these:
```
** (io.elementary.mail:3157): CRITICAL **: 11:31:11.615: mail_grouped_folder_source_item_build_folder_full_name: assertion 'folder_uri != NULL' failed

** (io.elementary.mail:3157): WARNING **: 11:31:11.615: GroupedFolderSourceItem.vala:134: Grouped folder '(null)' not available for account 'daniel@elementary.io'

** (io.elementary.mail:3157): CRITICAL **: 11:31:11.619: mail_grouped_folder_source_item_build_folder_full_name: assertion 'folder_uri != NULL' failed

** (io.elementary.mail:3157): WARNING **: 11:31:11.619: GroupedFolderSourceItem.vala:134: Grouped folder '(null)' not available for account 'daniel.p.fore@gmail.com'

(io.elementary.mail:3157): camel-WARNING **: 11:31:13.398: CamelIMAPXStore::get_folder_info_sync() reported failure without setting its GError

** (io.elementary.mail:3157): WARNING **: 11:31:13.398: GroupedFolderSourceItem.vala:134: Grouped folder '%5bGmail%5d/Drafts' not available for account 'daniel@elementary.io'

(io.elementary.mail:3157): camel-WARNING **: 11:31:14.585: CamelIMAPXStore::get_folder_info_sync() reported failure without setting its GError

** (io.elementary.mail:3157): WARNING **: 11:31:14.585: GroupedFolderSourceItem.vala:134: Grouped folder '%5bGmail%5d/Drafts' not available for account 'daniel.p.fore@gmail.com'
```

when clicking sent:
```
g_variant_new_string: assertion 'string != NULL' failed
```

danrabbit:
Another small thing I'm noticing is that when we first open Mail, it restores the selected folder for example if you select Inbox under one of your accounts. But if the last selected folder was one of these grouped folders, it does not restore on startup

marbetschar:
- Reverted to INBOX only
- Automatically select INBOX on app start
- Merged latest master

src/ConversationList/ConversationListBox.vala
```diff
@@ -253,20 +271,33 @@ public class Mail.ConversationListBox : VirtualizingListBox {
     }
 
     public async int archive_selected_messages () {
-        var threads = new Gee.ArrayList<Camel.FolderThreadNode?> ();
+        var archive_threads = new Gee.HashMap<string, Gee.ArrayList<Camel.FolderThreadNode?>> ();
         var selected_rows = get_selected_rows ();
-        foreach (unowned GLib.Object row in selected_rows) {
-            threads.add (((ConversationItemModel)row).node);
+        foreach (var selected_row in selected_rows) {
```
danrabbit, line 276: should this still be:
```suggestion
        foreach (unowned var selected_row in selected_rows) {
```

src/ConversationList/ConversationListBox.vala
```diff
@@ -276,13 +307,22 @@ public class Mail.ConversationListBox : VirtualizingListBox {
     }
 
     public int trash_selected_messages () {
-        var threads = new Gee.ArrayList<Camel.FolderThreadNode?> ();
+        var trash_threads = new Gee.HashMap<string, Gee.ArrayList<Camel.FolderThreadNode?>> ();
+
         var selected_rows = get_selected_rows ();
-        foreach (var row in selected_rows) {
-            threads.add (((ConversationItemModel)row).node);
+        foreach (var selected_row in selected_rows) {
```
danrabbit, line 313: unowned again probably?
```suggestion
        foreach (unowned var selected_row in selected_rows) {
```

danrabbit (commented):
This works as expected for me, nice job! Only two very small code comments

marbetschar:
@danrabbit made the fixes as suggested. Everything still works on my end, so I guess we are good. Still not 100% sure about when to use `unowned` though :)

danrabbit (approved):
LFG :rocket:

PR #568, opened 2021-05-04
Spelling: Send without subject

comradekingu:


cassidyjames (approved):
Yeah I like this title better and the description is a little more concise. :+1:

PR #573, opened 2021-05-14
CI: update flatpak-action

danrabbit:
* Only run CI on PRs
* Update flatpak action
* Make lint use checkout v2


btkostner:
I updated this to match what Camera has (plus also a build step in CI which might not be needed anymore). So now it includes publishing daily flatpak as well

cassidyjames (approved):

PR #585, opened 2021-06-02
MessageListItem: try to fetch Gravatar

### LINKED ISSUE #269: Try to fetch fallback avatars from Gravatar
If we can't get an avatar from Folks, we should try Gravatar

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/64898801-try-to-fetch-fallback-avatars-from-gravatar?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F59850425&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #269

Generates a gravatar URL and tries to set avatar images if available. Caches results so you only fetch once

This was a feature in Geary, but there is some concern about pinging a foreign service so I'm not sure if it needs some discussion

Technically the function of this program is to send data to a foreign service. I'm not sure if gravatar presents more of a risk surface than Gmail. But maybe this warrants a gsetting or something?


tintou:
I can see it being a gsettings but that can be added later on, we can also add libravatar.org to the supported services later.

cassidyjames (approved):
This seems to work well enough to merge and iterate. Something I'm curious about is that I still get a brief flash of the HdyAvatar background color right before the Gravatar loads in, even when cached. That might be something to look into later.

PR #384, opened 2019-06-07
Make shortcuts work in the composer window

danrabbit:

PR #588, opened 2021-06-03
Composer: Remove unused buttons and bad shortcuts

danrabbit:
Until we can get #384 fixed up, this at least makes the composer UI honest with what buttons and shortcuts are available and work


cassidyjames (changes requested):
Mostly seems good, but I don't think Remove Formatting belongs linked to the other formatting buttons; with all of the others, you can have any combination of them selected at a time, but the Remove Formatting deselects the others which makes it behave inconsistently. It makes sense to be next to them, but maybe not linked?

I guess it's like the old school actual radio buttons where pressing one might pop the others up, but I don't like that it looks like yet-another kind of formatting when it's not.

danrabbit:
@cassidyjames no problem. Fixed

cassidyjames (approved):
:+1:

PR #592, opened 2021-06-09
Flatpak: punch a hole for libsecret

### LINKED ISSUE #591: Added in Online Accounts: But still prompt for password
## What Happened

When Mail is packaged using Flatpak, it is not capable of retrieving the before configured secrets from Online Accounts. This leads to a password prompt when Mail is freshly started after adding a new account in System Settings -or as @danrabbit puts it in [his comment](https://github.com/elementary/switchboard-plug-onlineaccounts/pull/184#pullrequestreview-676513515):

> I'm still having trouble with the saved password in the flatpak, but this does seem to work for native mail. I wonder if there's a dbus interface we're missing that we need to add to the Mail sandbox

## Expected Behavior

Once the account is added in Online Accounts, it should be fully usable from Mail - without forcing the user to retype his password again.

## Steps to Reproduce

In the latest elementary OS Beta:

1. Add a new Mail account through `System Settings > Online Accounts`
2. Open Mail
3. A password prompt appears

## Potential solution

as @Marukesu [points out](https://github.com/elementary/switchboard-plug-onlineaccounts/pull/184#issuecomment-855288992), we are probably dealing with a secrets management issue in Flatpak-Land:

> eds use libsecret to read/save passwords, because we are on a sandbox, libsecret try to get them from a file on the sandbox instead of the DBus Interface.

Regarding [this](https://opensource.com/article/19/11/secrets-management-flatpak-applications) article, we might want fall back to the legacy behaviour by allowing access to the Systems Secrets Service API via DBus (`org.freedesktop.secrets`). But I'm wondering, if there is a more elegant solution to this matter, since this solution puts a hole in the Flatpak sandbox.

danrabbit:
Fixes #591 

Maybe not super ideal since we're busting a hole in the sandbox, but some sandboxing is better than none! We can always tighten up as we go

### DIFF
--- a/io.elementary.mail.yml
+++ b/io.elementary.mail.yml
@@ -21,7 +21,7 @@ finish-args:
 
   # LibSecret interfaces
   - '--talk-name=org.freedesktop.secrets'
-	- '--talk-name=org.gnome.keyring.SystemPrompter'
+  - '--talk-name=org.gnome.keyring.SystemPrompter'
 
   - '--metadata=X-DConf=migrate-path=/io/elementary/mail/'
 cleanup:
--- a/io.elementary.mail.yml
+++ b/io.elementary.mail.yml
@@ -19,6 +19,9 @@ finish-args:
   - '--talk-name=org.gnome.evolution.dataserver.AddressBook10'
   - '--talk-name=org.gnome.evolution.dataserver.Sources5'
 
+  # LibSecret interfaces
+  - '--talk-name=org.freedesktop.secrets'
+
   - '--metadata=X-DConf=migrate-path=/io/elementary/mail/'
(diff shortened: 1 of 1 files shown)


tintou (approved):

tintou:
That's the price to pay for desktop integration at least :slightly_smiling_face:

danrabbit:
It doesn't appear that this actually works

hanaral:
Do we know why this doesn't work? Mail and Epiphany would probably benefit the most from sandboxing

danrabbit:
@hanaral see discussion on the issue report: https://github.com/elementary/mail/issues/591#issuecomment-857967314

PR #598, opened 2021-06-25
ConversationListBox: Add a tooltip when hovering a message

tintou:
This allows to easily see the summary of a message

### DIFF
--- a/src/ConversationList/ConversationItemModel.vala
+++ b/src/ConversationList/ConversationItemModel.vala
@@ -161,8 +161,12 @@ public class Mail.ConversationItemModel : GLib.Object {
         return i;
     }
 
-    private static int64 get_newest_timestamp (Camel.FolderThreadNode node, int64 highest = -1) {
+    private static int64 get_newest_timestamp (Camel.FolderThreadNode? node, int64 highest = -1) {
         int64 time = highest;
+        if (node == null) {
+            return time;
+        }
+
         weak Camel.MessageInfo? message = node.message;
         if (message != null) {
             time = int64.max (time, message.date_received);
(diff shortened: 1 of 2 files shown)


marbetschar:
The code itself looks good, but as I ran it I noticed two things we can improve:

**1. The tooltips location is not updated while it is shown**

The code does update the tooltips' text, but not its location. This causes the tooltip to "hang" close to the message where it was rendered initially - but showing the text of the newly hovered message:

Screenshot from 2021-07-02 10-00-13

**2. Don't show any tooltip if the full text is shown already**

Probably self explanatory :)

Screenshot from 2021-07-02 10-02-50

tintou:
I've fixed suggestion 1, I can't do anything for suggestion 2 as I have no indication of which part of the label is actually drawn or not.

src/ConversationList/ConversationListBox.vala
```diff
@@ -348,4 +349,15 @@ public class Mail.ConversationListBox : VirtualizingListBox {
...
+    public override bool query_tooltip (int x, int y, bool keyboard_tooltip, Gtk.Tooltip tooltip) {
+        unowned VirtualizingListBoxRow? row = get_row_at_y (y);
+        if (row != null) {
+            unowned ConversationItemModel model_item = (ConversationItemModel) row.model_item;
+            tooltip.set_markup (GLib.Markup.printf_escaped ("<b>%s</b>\n%s", model_item.from, model_item.subject));
+            return true;
+        }
+
+        return false;
+    }
```
marbetschar, line 362: Since you set the `tooltip_markup` in `ConversationListItem` now, I guess we can completely remove this handler?

marbetschar (changes requested):

src/ConversationList/ConversationListBox.vala
tintou: It was on my local copy, I don't know what happened on GitHub side :man_shrugging:

marbetschar: Had this already once in the past - I guess the code was re-added on my side somehow related to your force push. Deleted my local branch and pulled it again completely fresh. Looks good now 👍

marbetschar (approved):

Issue #340, opened 2018-12-31
Mail not saving drafts


AbhinavMadahar:
Pantheon Mail does not save my drafts after I write them. I'm using Elementary OS 5.0. This happens when the "Save drafts on server" option is enabled and when it's disabled.

casasfernando:
I have exactly the same problem and it's really annoying.

PR #599, opened 2021-07-01
Save Draft

marbetschar:
This PR fixes https://github.com/elementary/mail/issues/340. However, there are a few caveats with the current code base:

**1. We are saving to drafts folder while the mail composer is destroyed**

~~This implies that we are not able to use any async operations - which in turn causes the UI to block for a brief moment. If we don't want this we need to figure out another way/event when to save a draft, which probably needs some more design work.~~

Async saving draft is possible during destroy if we override the default destroy handler - which is the way it is implemented now. This solves the issue of the blocking UI.

**2. There is no UI for editing a before saved draft**

While drafts are saved now, there is no way of editing a before saved draft. This needs some more design work as well.


danrabbit:
It's probably fine to save draft editing for another branch since you could theoretically edit saved drafts from another client or copy/paste at worst. One step at a time :)

marbetschar:
@danrabbit alright then, I guess this is ready for review!

src/Backend/Session.vala
```diff
@@ -363,4 +382,69 @@ public class Mail.Backend.Session : Camel.Session {
...
+            }
+        }
+
+        return null;
+    }
+
+    public async void save_draft (Camel.MimeMessage message, Camel.InternetAddress from, Camel.Address recipients) throws Error {
+        var camel_store = get_camel_store_from_email (from);
+        if (camel_store == null) {
+            warning ("No camel store for saving found, discarding draft.");
```
danrabbit, line 428: shouldn't we throw an error here?

src/Backend/Session.vala
```diff
@@ -363,4 +382,69 @@ public class Mail.Backend.Session : Camel.Session {
...
+            return;
+        }
+
+        var drafts_folder_uri = get_drafts_folder_uri_for_store (camel_store);
+        if (drafts_folder_uri == null || drafts_folder_uri == "") {
+            warning ("Unable to determine drafts folder uri, discarding draft.");
```
danrabbit, line 434: and throw an error here as well?

src/Composer/ComposerWidget.vala
```diff
@@ -787,4 +831,41 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+    public override void destroy () {
+        if (discard_draft) {
+            base.destroy ();
+            return;
+        }
+
+        web_view.get_body_html.begin ((obj, res) => {
+            var body_html = web_view.get_body_html.end (res);
+
+            if (body_html == null || get_first_line_without_html (body_html) == first_line_of_message_without_html_on_load) {
```
danrabbit, line 844: Can you explain what this is checking for? I don't understand what's going on with the whole `first_line_of_message_without_html` and what the significance is of that

src/Composer/ComposerWidget.vala
```diff
@@ -787,4 +831,41 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+                session.save_draft.begin (
+                    message,
+                    sender,
+                    recipients,
+                    (obj, res) => {
+                        try {
+                            session.save_draft.end (res);
+                        } catch (Error e) {
+                            critical ("Unable to safe draft: %s", e.message);
+                            // TODO: Shall we display a dialog here?
```
danrabbit, line 863: Yeah this should definitely not silently fail and in fact we should probably not quit the composer if this returns an error to avoid data loss

src/Utils.vala
```diff
@@ -61,4 +61,22 @@ public class Mail.Utils {
     public static string escape_html_tags (string input) {
         return input.replace ("<", "&lt;").replace (">", "&gt;");
     }
+
+    public static string strip_html_tags (string input) {
```
danrabbit, line 65: If this is only used in one place, it probably doesn't need to be its own public util in another class

src/Utils.vala
```diff
@@ -61,4 +61,22 @@ public class Mail.Utils {
...
+
+        } catch (RegexError e) {
+            warning ("Error compiling regex to strip html tags: %s", e.message);
+            return input;
+        }
+    }
+
+    public static string? strip_folder_full_name (string service_uid, string? folder_uri) {
+        if (folder_uri != null) {
+            return folder_uri.replace ("folder://%s/".printf (service_uid), "");
```
danrabbit, line 78: I wonder if we also need to unescape here. I get an error about my Gmail drafts folder since the path is `[Gmail]/Drafts`:

`Unable to safe draft: No such folder %5bGmail%5d/Drafts`

danrabbit (changes requested):
Nice work on this! I left a few inline comments :)

src/Backend/Session.vala
marbetschar: yeah, good point!

src/Backend/Session.vala
marbetschar: yes!

src/Composer/ComposerWidget.vala
marbetschar: I tried to implement a way to detect whether the user changed anything or not. In case of a new draft this worked - but I removed the logic completely now, since thinking about it a bit more it will no longer work once we are able to edit drafts.

src/Composer/ComposerWidget.vala
marbetschar: I implemented a re-rendering of the ComposerWidget within a new ComposerWindow in case saving the draft fails. In addition, an error dialog is shown now.

src/Utils.vala
marbetschar: Removed completely, since its no longer used.

src/Utils.vala
marbetschar: Added Camel.URL.decode when retrieving the drafts folder.

marbetschar:
@danrabbit thank you very much for the review - lots of good points in there! I addressed all of them and the archive folder uri is now decoded. Can you give it a re-test with your Gmail account?

src/Composer/ComposerWidget.vala
```diff
@@ -787,4 +808,63 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+                            }
+
+                            if (main_window != null) {
+                                new ComposerWindow.for_widget (main_window, this).show_all ();
+                            } else {
+                                warning ("Unable to re-show composer. Draft will be lost.");
+                            }
+
+                            var error_dialog = new Granite.MessageDialog (
+                                _("Unable to safe draft"),
```
danrabbit, line 857:
```suggestion
                                _("Unable to save draft"),
```

danrabbit (changes requested):
I can confirm that draft saving now works here and error handling works as well. Nice work!

It looks like we're also saving a draft when sending though :upside_down_face:

marbetschar:
> It looks like we're also saving a draft when sending though 🙃

*haha*, yeah - thats fixed now :)

danrabbit (approved):
Nice! Let's go :rocket:

PR #602, opened 2021-07-06
Set and show subject on Reply

### LINKED ISSUE #601: Reply Subject not set
## What Happened

When I reply to an email, the subject does not get set and there is also no subject field rendered. This leads to the "Send without subject?" error message to be shown as soon as I hit the send button.

## Expected Behavior

When I reply to an email, I'd expect the subject to be set. In addtion, the subject should be rendered in a text entry, so I'm able to edit it as needed before sending the email.

## Steps to Reproduce

1. Select an email from you inbox
2. click the Reply icon
3. write your reply
4. click the send button
5. the following error message appears

Screenshot from 2021-07-04 22-07-36

## Platform Information

Latest elementary 6 Odin (daily).

marbetschar:
This PR makes sure the subject is set and shown when replying to an email. Fixes #601.

### DIFF
--- a/src/Composer/InlineComposer.vala
+++ b/src/Composer/InlineComposer.vala
@@ -36,6 +36,9 @@ public class Mail.InlineComposer : Gtk.ListBoxRow {
         composer = new ComposerWidget.inline ();
         composer.margin_top = 6;
         composer.has_recipients = true;
+        if (prev_chain_message != null && prev_chain_message.get_subject () != null) {
+            composer.subject = _("Re: %s").printf (prev_chain_message.get_subject ());
+        }
         composer.discarded.connect (() => {
             discarded ();
         });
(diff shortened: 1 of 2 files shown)


src/Composer/InlineComposer.vala
```diff
@@ -36,6 +36,9 @@ public class Mail.InlineComposer : Gtk.ListBoxRow {
         composer = new ComposerWidget.inline ();
         composer.margin_top = 6;
         composer.has_recipients = true;
+        if (prev_chain_message != null && prev_chain_message.get_subject () != null) {
+            composer.subject = _("Re: %s").printf (prev_chain_message.get_subject ());
```
tintou, line 40: The [email RFC](https://datatracker.ietf.org/doc/html/rfc2822#section-3.6.5) makes this string verbatim (so don't translate it) also we have to make sure that the subject doesn't already contain it before re-adding it.

src/Composer/InlineComposer.vala
```diff
@@ -36,6 +36,9 @@ public class Mail.InlineComposer : Gtk.ListBoxRow {
...
+        }
```
tintou, line 41: empty line after closing brace

tintou (approved):

PR #603, opened 2021-07-08
Handle subject in ComposerWidget

marbetschar:
This makes the code more straight forward and we can get rid of some state handling.


davidmhewitt:
Just opened a semi-related issue at #604

marbetschar:
I can certainly add case insensitivity to this PR.

In terms of Microsoft Outlook unfortunately things are even worse: Outlook localizes "Re:" and "FW" - for instance in German, "AW:" and "WG:" are added. So I don't think there is anything we can do here unless we want to open pandoras box and support all non-standard-conforming mail clients out there.

davidmhewitt:
> Outlook localizes "Re:" and "Fwd:" - for instance in German, "AW:" and "WG:" are added. So I don't think there is anything we can do here unless we want to open pandoras box and support all non-standard-conforming mail clients out there.

Ah yeah, I'd forgotten about that. But that's definitely where I've seen silly long chains of these before when I've worked with German companies at my day job, "AW: RE: AW: RE: AW: RE: " 😆 

There's actually an option in Outlook to make it always use the non-localised version for those, but it's not enabled by default. Funny how the default option is "Ignore the email RFC".

I think supporting case-insensitive "Re:" and "Fwd:" and "Fw:" at least prevents it happening in a few of the common cases, and people that email people with misbehaving mail clients will probably already be used to seeing it happen in the other cases.

src/Composer/ComposerWidget.vala
```diff
@@ -480,6 +478,28 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
+                // https://datatracker.ietf.org/doc/html/rfc2822#section-3.6.5
+                if (!subject_val.text.up ().contains ("RE: ")) {
+                    subject_val.text = "Re: %s".printf (subject_val.text);
+                }
+
+            } else if (type == Type.FORWARD) {
+                // Although "Fwd: " is not part of the RFC standard,
+                // it is a widely used convention - therefore we stick
+                // to it here too. Do _not_ translate it!
+                if (!subject_val.text.up ().contains ("FWD: ") && !subject_val.text.up ().contains ("FW: ")) {
```
tintou, line 497:
```suggestion
                var up_subject = subject_val.text.up ()
                if (!up_subject.contains ("FWD: ") && !up_subject.contains ("FW: ")) {
```

tintou (approved):

PR #605, opened 2021-07-08
Add Drafts icons

danrabbit:
Thought it would be nice to use a badged icon for dialogs here and I noticed that we don't have a 48px drafts icon and we were actually using folder-documents for drafts. So gresource some specific icons

Screenshot from 2021-07-08 10 28 36


marbetschar (changes requested):
The changes look good - just one missing piece: We probably want to change the icon in [FolderSourceItem:82](https://github.com/elementary/mail/blob/4ffd94e31b3cddfb2fd29284b85a4f9bebb6c644/src/FoldersView/FolderSourceItem.vala#L82) as well.

marbetschar (approved):

Issue #550, opened 2021-03-02
Sent emails are not added to "Sent" folder


cassidyjames:
This might be a duplicate/related to #468, but any emails I'm sending are not being added to the Sent folder on my account as far as I can tell.

marbetschar:
Same here: Message is not saved in Sent. This is probably just a matter of saving a copy of the mail in the Sent folder - similar to how it is implemented for saving a draft in https://github.com/elementary/mail/pull/599

And while we are at it: Error handling of sending an email can be improved. Right now it does not provide any user facing feedback if it fails and the mail is just discarded in this case. The error handling is implemented in draft saving too.

Issue #468, opened 2020-04-10
Sent doesn't sync to iCloud


evennotodd:
Mail sent from the client doesn't sync to my iCloud sent folder, but it successfully sends. Is there any way to get it to sync and fix this?

evennotodd:
I will say that I can manually move them to the iCloud sent folder that is under labels and it works. Is there any way to make this move automatic at the least?

PR #607, opened 2021-07-09
Send sent to Sent

marbetschar:
This PR fixes https://github.com/elementary/mail/issues/550 and https://github.com/elementary/mail/issues/468. The idea is to re-use some code introduced in https://github.com/elementary/mail/pull/599 - so we need to get the Save Draft pull request merged first before I can continue to work on this.


danrabbit (approved):
I can can confirm that this works as intended when the sent mailbox is set correctly. It warns when it is not set and errors when the mail can't be send.

Nice work!

PR #600, opened 2021-07-02
Fix Warnings in Vala Code

marbetschar:
This PR fixes all the warnings in the Vala codebase - making it a bit easier to work on new stuff.

PR #622, opened 2021-07-22
meson: require libhandy-1 >= 1.1.90

decathorpe:
The Hdy.Avatar loadable-icon property was added in libhandy 1.1.90:
https://gitlab.gnome.org/GNOME/libhandy/-/blob/master/NEWS

Usage of this API was added in PR #600  / commit https://github.com/elementary/mail/commit/9477c24 , which claimed to only fix Vala warnings, so this was probably an oversight.


tintou (approved):

tintou:
We need to update the README too

Issue #623, opened 2021-07-22
Office 365: Sent stored twice


marbetschar:
## What Happened

Since https://github.com/elementary/mail/pull/607 sent mails are copied to the Sent folder. However, in Office 365 the mail appears twice in the Sent folder (Screenshot of Sent folder in Office 365 web interface):

![Sent-Stored-Twice](https://user-images.githubusercontent.com/392542/126655037-37828178-ae30-433d-bda7-65e94474be2e.PNG)

## Expected Behavior

The sent mail should only be stored once in the Sent folder.

## Steps to Reproduce

1. Configure your Office 365 mail account in System Settings > Online Acconts
2. Start Evolution and let it detect all well known folders (should be fixed once https://github.com/elementary/switchboard-plug-onlineaccounts/pull/192 is merged)
3. Close Evolution
4. Start elementary Mail and verify the Sent folder is rendered using the proper icon
5. Send an email from your Office 365 account
6. You'll notice the sent mail will be stored twice in your Office 365 account

## Platform Information

elementary OS 6 Daily.

PR #625, opened 2021-07-23
Don't store two copies in Sent for Office 365 and Google

marbetschar:
Fixes https://github.com/elementary/mail/issues/623 for Office 365 and Google.


danrabbit (approved):
I can't confirm that I get two messages in sent for Gmail, but it seems like it still does one here, so :+1:

Issue #610, opened 2021-07-13
Edit and Send Draft


marbetschar:
## Problem

Even though drafts are saved now, it is still not possible to edit a previously saved draft and send it once it is finished.

Issue #619, opened 2021-07-20
Don't attempt to save draft if empty


kdwk:
## What Happened

https://user-images.githubusercontent.com/39268094/126264023-14ce084f-d010-4ada-8fa4-92a083cfdd97.mp4

Mail attempts to save the draft even when there is nothing in the draft

## Expected Behavior

If there is nothing in the email yet, don't try to save it as a draft

## Steps to Reproduce

1. Compose mail
2. Don't add anything
3. Close the compose mail window
4. It will try to save it to drafts

## Platform Information

![Screenshot from 2021-07-14 10-55-34](https://user-images.githubusercontent.com/39268094/126264186-0eb230a3-a154-46de-b628-190465833879.png)

Mail 6.0.0

marbetschar:
@davidmhewitt are you aware of an easy way to detect changes in the composer or the webview after initial load?

PR #626, opened 2021-07-23
Edit draft

marbetschar:
Fixes https://github.com/elementary/mail/issues/610

### DIFF
--- a/src/Backend/Session.vala
+++ b/src/Backend/Session.vala
@@ -472,7 +472,7 @@ public class Mail.Backend.Session : Camel.Session {
         return null;
     }
 
-    public async void save_draft (Camel.MimeMessage message, Camel.InternetAddress from, Camel.Address recipients, string? ancestor_message_info_uid = null) throws Error {
+    public async void save_draft (Camel.MimeMessage message, Camel.InternetAddress from, Camel.Address recipients, Camel.MessageInfo? ancestor_message_info = null) throws Error {
         var camel_store = get_camel_store_from_email (from);
         if (camel_store == null) {
             throw new Camel.ServiceError.UNAVAILABLE ("No camel service for saving draft found.");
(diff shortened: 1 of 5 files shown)


src/Backend/Session.vala
```diff
@@ -472,7 +472,7 @@ public class Mail.Backend.Session : Camel.Session {
         return null;
     }
 
-    public async void save_draft (Camel.MimeMessage message, Camel.InternetAddress from, Camel.Address recipients) throws Error {
+    public async void save_draft (Camel.MimeMessage message, Camel.InternetAddress from, Camel.Address recipients, string? ancestor_message_info_uid = null) throws Error {
```
tintou, line 475:
```suggestion
    public async void save_draft (Camel.MimeMessage message, Camel.InternetAddress from, Camel.Address recipients, Camel.MessageInfo? ancestor_message_info = null) throws Error {
```
Any reason we can't directly use the right messageinfo object here? (this would avoid a lookup)

marbetschar: @tintou good catch! Passing `Camel.MessageInfo` directly works as expected.

tintou (approved):

danrabbit (commented):
Hm, I'm not sure I'm understanding what this is intended to do

When I go to my drafts folder and select a message, it looks like there's some kind of editable text view here now, but there's no way to send the message or save my edits as far as I can tell

Selecting a message should probably open it with the inline composer right?

marbetschar:
@danrabbit exactly: the message is rendered and the inline composer is shown right below. There you are capable to edit/send etc.

Isn't that the case on your end?

danrabbit (changes requested):
Ah, okay going through the list of drafts I'm getting it for some emails and not for others. I'm not sure why.

I think rendering the message first is confusing. It's not clear that you need to scroll down to get to the draft and it looks like the message has already been sent as part of a conversation that way

I'm also noticing that none of the drafts preserve recipient information, so if you have a draft with a recipient saved and you open it, the "To" entry is blank and then when it saves it'll remove the recipient line

danrabbit:
It also looks like we're saving the draft even if no changes have been made, so the timestamps are getting bumped to today for old drafts

marbetschar:
@danrabbit 

> Ah, okay going through the list of drafts I'm getting it for some emails and not for others. I'm not sure why.

Did you create those drafts in elementary Mail? If so, then the [`Camel.MessageFlags.DRAFT` is probably not set on the message](https://github.com/elementary/mail/blob/edit-draft/src/MainWindow.vala#L199). In my tests, I was capable to create a draft on the server side and then successfully edit it in Mail. Also if you create a new draft while testing this branch, the flag gets set correctly - so that should work too.

> I think rendering the message first is confusing. It's not clear that you need to scroll down to get to the draft and it looks like the message has already been sent as part of a conversation that way

Yes, I totally agree. However, this was the easiest way to add, since the code in place (the composer) expects a conversation to be set to get the content to cite from. I now added code which scrolls to the bottom once the inline composer was added for editing a draft.

In terms of the overall message/replying rendering handling, I started [a discussion here](https://github.com/elementary/mail/discussions/609).

> I'm also noticing that none of the drafts preserve recipient information, so if you have a draft with a recipient saved and you open it, the "To" entry is blank and then when it saves it'll remove the recipient line

Good catch!! Somehow missed that one - is now fixed in the latest commit.

> It also looks like we're saving the draft even if no changes have been made, so the timestamps are getting bumped to today for old drafts

This is an expected side effect as of now. This problem needs also to be solved when creating a new draft - but not adding any content as described here: https://github.com/elementary/mail/issues/619

danrabbit (approved):
Thanks for explaining!

I can confirm the fix for the recipient and scrolling down to the composer solves my main issue there

Happy to merge and continue iterating on other stuff. Thanks!

PR #627, opened 2021-07-25
Fix MailBackendSimpleSasl has an empty CamelServiceAuthType

marbetschar:
This PR attempts to fix the critical message printed at the startup of Mail:

```
(io.elementary.mail:6659): camel-CRITICAL **: 11:20:27.758: MailBackendSimpleSasl has an empty CamelServiceAuthType
```

Which is caused by the following line in EDS: https://gitlab.gnome.org/GNOME/evolution-data-server/-/blob/master/src/camel/camel-sasl.c#L103

However, I'm stuck: Because I can't subclass `Camel.ServiceAuthType`, I simply cast the newly added `SimpleServiceAuthType` to `Camel.ServiceAuthType` which works just fine. However, EDS is still throwing the critical message. @tintou what I'm missing here?

### DIFF
--- a/src/Backend/Session.vala
+++ b/src/Backend/Session.vala
@@ -19,22 +19,19 @@
  */
 
 public class Mail.Backend.SimpleSasl : Camel.Sasl {
-    [CCode (cname = "auth_type")]
-    public Camel.ServiceAuthType auth_type;
+    private static Camel.ServiceAuthType simple_auth_type = {
+        N_("SimpleServiceAuthType"),
+        N_("Provides a way to fall back to passwords in SSO scenarios."),
+        "SIMPLESERVICEAUTHTYPE",
+        false
+    };
 
     public SimpleSasl (string service_name, string mechanism, Camel.Service service) {
         Object (service_name: service_name, mechanism: mechanism, service: service);
     }
 
-    construct {
-        auth_type = (Camel.ServiceAuthType) new SimpleServiceAuthType ();
-    }
-
-    private class SimpleServiceAuthType {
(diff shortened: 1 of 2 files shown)


tintou:
@marbetschar I've fixed the introspection, I haven't actually properly tested it though

tintou:
@marbetschar I've fixed the introspection, I haven't actually properly tested it though

tintou:
@marbetschar I've fixed the introspection, I haven't actually properly tested it though

tintou:
@marbetschar I've fixed the introspection, I haven't actually properly tested it though

marbetschar:
@tintou wow, that's impressive! Thank you so much!! That change would have taken me ages to figure out :laughing:

Things are working flawlessly on my end - not quite sure what you mean with "properly" tested though. I guess we should hand this in for review and see what others think.

marbetschar:
⚠ WARNING: Some changes introduced here caused a SegmentationFault in https://github.com/elementary/mail/pull/679#issuecomment-913172123

No clue why exactly 🤷‍♂️

tintou:
> warning WARNING: Some changes introduced here caused a SegmentationFault in [#679 (comment)](https://github.com/elementary/mail/pull/679#issuecomment-913172123)
> 
> No clue why exactly man_shrugging

We need a backtrace with the symbols enabled to understand why it's the case

marbetschar:
@tintou which simbols are you missing from [the stacktrace linked](https://github.com/elementary/mail/pull/679#issuecomment-913172123)?

PR #634, opened 2021-08-10
Set the correct from address on reply in multi account environments

marbetschar:
This change makes sure Mail sets the correct from address when multiple accounts are configured. Before this change, Mail just takes the first address to respond to any mail - regardless in which inbox the mail was originally received.

This PR also makes sure we always show the from address field when replying or forwarding an email in case multiple accounts are configured. This allows the user to change the sender account when needed.


marbetschar:
Fixes https://github.com/elementary/mail/issues/658

tintou (approved):

janxkoci:
Interesting, this was supposed to be fixed months ago, but it just happened to me today that I replied to an email from a wrong account. I have the latest version (6.4 according to AppCenter), as I upgraded my machine to 6.1 only last week (fresh reinstall from OS 5.1). So how is this possible?

PR #648, opened 2021-08-16
AppData: Use Banana 300 for brand banner

cassidyjames:
Plus black foreground text. I experimented with a really dark "yellow" for the foreground but it looked like an ugly brown to me. Not totally opposed to the idea if we can get something nice.

Screenshot from 2021-08-16 14-16-52

Screenshot from 2021-08-16 14-15-27


danrabbit (approved):
Hot

PR #655, opened 2021-08-22
Fix #582: Calculate the page height correctly

### LINKED ISSUE #582: Certain Mails are not rendered
## What Happened

Certain Emails are not rendered (displayed) in elementary Mail (elementary OS Odin). These _certain mails_ which are not rendered are visible when the reply button is clicked on the mail. I feel there is some problem in the code which displays mail and not the email because, the email is correctly visible when the reply button is clicked.

## Expected Behavior

The normal way mails are displayed is the behavior expected from elementary Mail.

## Steps to Reproduce

There are no steps to reproduce. Certain Mails are not displayed.

Adding a few screenshots below to best describe the issue:

Mail 1 (Displayed Properly)**

Mail 2 (Not displayed properly)**
01-same

Mail 2 (The same email, when the reply button is clicked)**

****Text has been blurred for privacy concerns**

### Logs

No Logs

## Platform Information

Operating System: elementary OS 6
Edition: (Public Beta)
Hardware Information:
- Quad-Core Intel® Core™ i5-8250U CPU @ 1.60GHz
- Intel® UHD Graphics 620 (Kabylake GT2)
- 7.7 GiB memory
- 467.2 GB storage (SSD)

Laptop Information: HP Pavilion x360

marbetschar:
This PR fixes #582 by correctly calculating the page height of an email as the previously used method failed under certain circumstances.

We now just use the max height available. Solution inspired by this StackOverflow post: https://stackoverflow.com/questions/1145850/how-to-get-height-of-entire-document-with-javascript#answer-1147768

### DIFF
--- a/webkit-extension/MailPage.vala
+++ b/webkit-extension/MailPage.vala
@@ -46,7 +46,12 @@ public class Mail.Page : Object {
                 message.send_reply (new WebKit.UserMessage ("get-body-html", new Variant.take_string (val.to_string ())));
                 return true;
             case "get-page-height":
-                JSC.Value val = js_context.evaluate ("document.documentElement.offsetHeight;", -1);
+                JSC.Value val = js_context.evaluate ("""
+                Math.max(
+                    document.body.scrollHeight, document.body.offsetHeight,
+                    document.documentElement.clientHeight, document.documentElement.scrollHeight, document.documentElement.offsetHeight
+                );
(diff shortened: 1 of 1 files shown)


davidmhewitt (approved):
Can confirm that this fixes the issue for an email I have here. Doesn't seem to negatively affect a bunch of other emails I've opened.

PR #656, opened 2021-08-22
Fix #674: Cannot delete emails

marbetschar:
We need to unfreeze the folder after we are done with the delete operation to propagate the folder change. The reason why some messages seem to be deleted before this fix was due to the undo timeout kicking in after 10 seconds and doing exactly this.

This PR does the unfreeze immediately after deletion, thus - even if Mail is closed before the 10 seconds timeout - deleted messages do not come back from the dead.

However: The current implementation is troublesome, because it sets the delete flag of the message and there is way to recover a deleted message once the undo timeout ran down. A better way of doing this would be not to set the delete flag, but rather move the message to the [Trash folder](https://valadoc.org/camel-1.2/Camel.Store.get_trash_folder.html) and let the server decide when to explicitly delete it. Will open another issue for this.

### DIFF
--- a/src/Backend/MoveHandler.vala
+++ b/src/Backend/MoveHandler.vala
@@ -26,12 +26,12 @@ public class Mail.MoveHandler {
     private Camel.Folder src_folder;
     private Camel.Folder? dst_folder;
     private Gee.ArrayList<weak Camel.MessageInfo> moved_messages;
-    private bool frozen = false;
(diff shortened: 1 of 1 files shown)


marbetschar:
The follow up issue for moving deleted mails to trash folder instead of setting the delete flag https://github.com/elementary/mail/issues/657

PR #660, opened 2021-08-23
Automatically select next message after delete.

### LINKED ISSUE #641: When deleting mail via delete key, the next message does not get automatically selected - Quick delete not possible
## What Happened

When selecting an email in the list of messages, pressing `Delete` will delete the message, but the next message in the inbox/list does not get selected, which means `Delete` cannot be pressed again to just delete the following message, and so on.

## Expected Behavior

When `Delete` is pressed, the message should get trashed/deleted, and the next message in the inbox should be selected/active to allow the user to act on it.

## Steps to Reproduce

1. Select an email in the inbox
2. Press `Delete`
3. Attempt to delete again for the next message without use of mouse

## Platform Information

elementary OS 6.0 Odin - Fully updated as of 08/12/2021 05:00

marbetschar:
Fixes #641.

@davidmhewitt had to overwrite `get_index_of` in `ConversationListStore`: Is there a specific reason why `RowVisibilityFunc` is not part of `VirtualizingListBoxModel`? Because `get_index_of` and `get_index_of_unfiltered` are defined in `VirtualizingListBoxModel`.


marbetschar:
@davidmhewitt because I think you are familiar `VirtualizingListBox` used in Mail: Do you see any problems with the proposed implementation of automatically selecting the "next" message in this PR?

davidmhewitt (approved):
@marbetschar It does look like the `RowVisibilityFunc` should be a member of `VirtualizingListBoxModel` instead of the more specific class, but it's a long time since I wrote this so I don't remember why it was done that way. To be honest though, it will be nice to throw this away and use the GTK4 list model stuff in the future, so I'm not too worried.

I'm happy with your code changes here, but I haven't done any testing and I guess there's a potential for regressions with the overridden `get_index_of` method, so I'll let you decide whether to merge or wait for some testing by someone else.

marbetschar:
@davidmhewitt thanks for the review! I agree, we should use GTK4 for this as soon as possible.

I gave this a re-test and found a small bug with the current implementation: If we scroll far enough down (to older conversations), the code is no longer able to automatically select the next message after delete. Probably because we end up with a [higher message-index than widgets beeing actually rendered](https://github.com/elementary/mail/blob/16f735160d7713a1a36ba6e95918a47b934f4627/src/VirtualizingListBox/VirtualizingListBox.vala#L708). However, in this case the code falls back to the previous behaviour and simply does nothing. Unfortunately I don't see an easy fix because there seems no "offset" value available which would help in calculation.

So I'd argue we are still fixing https://github.com/elementary/mail/issues/641 - because if you want to mass-delete, changes are high the messages are relatively new and therefore somewhere near the top. So we gain some convenience here and don't have to wait for GTK4 to do this "properly".

Ultimately, I'll let @danrabbit decide if we want to merge this in :)

marbetschar:
Actually, l was wrong. I found a way to select the next message even when rows are virtualized. The way it works now allows us to get rid of the overwritten `get_index_of` as well.

@davidmhewitt any chance you can give this a re-review?

davidmhewitt (approved):
Code looks even better now, thanks for spending the time to look at it!

Still haven't done any testing, but this has much less chance of regressions now. So I'm happy

danrabbit (approved):
Mail is now complete! Great job ;p

Issue #652, opened 2021-08-20
Mail crashes after a few seconds


lfuelling:
## What Happened
I added my account, opened mail and waited for the messages to load and then browsed around a few messages to admire the messed up kerning in HTML mails. 

Then suddenly the program crashed.

I restarted the application by executing `io.elementary,mail` in a terminal and was able to get the log you can find below by using the same "browsing around" method as described above.

## Expected Behavior
Mail should keep running.

## Steps to Reproduce
1. Add an account with a huge mailbox and some large messages
2. Open Mail
3. Browse around for some time 
4. within two minutes the application should crash

### Logs
```
$ io.elementary.mail                                                                                                                                      

(io.elementary.mail:35588): camel-CRITICAL **: 20:27:13.395: MailBackendSimpleSasl has an empty CamelServiceAuthType

(io.elementary.mail:35588): camel-CRITICAL **: 20:27:13.395: MailBackendSimpleSasl has an empty CamelServiceAuthType

(io.elementary.mail:35588): Handy-WARNING **: 20:27:25.881: Failed to load icon: Not Found

** (process:2): WARNING **: 20:27:25.987: Error writing credentials to socket: Error sending message: Broken pipe

(io.elementary.mail:35588): Handy-WARNING **: 20:27:38.232: Failed to load icon: Not Found

(io.elementary.mail:35588): Handy-WARNING **: 20:27:38.233: Failed to load icon: Not Found

(io.elementary.mail:35588): Handy-WARNING **: 20:27:38.233: Failed to load icon: Not Found

(io.elementary.mail:35588): Handy-WARNING **: 20:27:47.040: Failed to load icon: Not Found

(io.elementary.mail:35588): Gdk-ERROR **: 20:27:49.287: The program 'io.elementary.mail' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadAlloc (insufficient resources for operation)'.
  (Details: serial 229010 error_code 11 request_code 53 (core protocol) minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the GDK_SYNCHRONIZE environment
   variable to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
``` 

## Platform Information
elementaryOS 6 Odin
Linux 5.13.0-12.1-liquorix-amd64

marbetschar:
Hi @lfuelling,

[...]

Issue #649, opened 2021-08-17
SIGSEGV Seg fault after comparing timestamps


RMcNeely:
## What Happened

Mail crashes after trying to compare timestamps

## Expected Behavior

Not crash

## Steps to Reproduce

1. Compiled Mail from release 6.1.0, following instructions in README
2. Started via CLI `io.elementary.mail`
3. Appears to happen after clicking on emails after a short time.

### Logs

bt from systemd-coredump
```
[build](6.1.0) sudo coredumpctl debug /usr/bin/io.elementary.mail
           PID: 94079 (io.elementary.m)
           UID: 1000 (ryan)
           GID: 1000 (ryan)
        Signal: 11 (SEGV)
     Timestamp: Tue 2021-08-17 12:16:11 EDT (35s ago)
  Command Line: io.elementary.mail
    Executable: /usr/bin/io.elementary.mail
 Control Group: /user.slice/user-1000.slice/session-c2.scope
          Unit: session-c2.scope
         Slice: user-1000.slice
       Session: c2
     Owner UID: 1000 (ryan)
       Boot ID: ed6d35659d59435488ba88d3e9efa4ae
    Machine ID: 049479c2daaf4a18868c383f62f96c41
      Hostname: ThinkPad-X1-Carbon-7th-e4472553
       Storage: /var/lib/systemd/coredump/core.io\x2eelementary\x2em.1000.ed6d35659d59435488ba88d3e9efa4ae.94079.1629216971000000000000.lz4
       Message: Process 94079 (io.elementary.m) of user 1000 dumped core.
                
                Stack trace of thread 94079:
                #0  0x00007fee733ae428 camel_message_info_get_date_received (libcamel-1.2.so.62 + 0x81428)
                #1  0x0000563a2d7a3c1b mail_conversation_item_model_get_newest_timestamp (io.elementary.mail + 0x4ac1b)
                #2  0x0000563a2d7a3d03 mail_conversation_item_model_get_newest_timestamp (io.elementary.mail + 0x4ad03)
                #3  0x0000563a2d7a511a mail_conversation_item_model_get_timestamp (io.elementary.mail + 0x4c11a)
                #4  0x0000563a2d7a8cca mail_conversation_list_box_thread_sort_function (io.elementary.mail + 0x4fcca)
                #5  0x0000563a2d7aafa8 _mail_conversation_list_box_thread_sort_function_gcompare_data_func (io.elementary.mail + 0x51fa8)
[...]

PR #664, opened 2021-08-29
Cache timestamp to avoid compare_func from crashing

dbhowell:
Avoids using freed nodes when `ConversationListStore` uses it's `compare_func`.

Fixed the `camel_message_info_get_date_received: assertion 'CAMEL_IS_MESSAGE_INFO (mi)' failed` errors for me (#649, #650 , #652).

Note: `folder_changed` updates *every* conversation in the folder, so is expensive when a folder has a large number of items. So for example, setting a message as read will ultimately mean `folder_changed` fires and touches every conversation. I can't see a way around it at the moment though.

### DIFF
--- a/src/ConversationList/ConversationItemModel.vala
+++ b/src/ConversationList/ConversationItemModel.vala
@@ -24,6 +24,8 @@ public class Mail.ConversationItemModel : GLib.Object {
     public string service_uid { get; construct; }
     public Camel.FolderThreadNode? node;
 
+    private int64 _timestamp;
+
     public string formatted_date {
         owned get {
             return Granite.DateTime.get_relative_datetime (new DateTime.from_unix_local (timestamp));
@@ -153,7 +155,7 @@ public class Mail.ConversationItemModel : GLib.Object {
 
     public int64 timestamp {
         get {
-            return get_newest_timestamp (node);
+            return _timestamp;
         }
     }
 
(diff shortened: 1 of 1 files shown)


marbetschar:
Neat, thank you @dbhowell!! @RMcNeely / @wout / @casasfernando: Can anyone of you guys confirm that this fix prevents Mail from crashing for you as well?

casasfernando:
I will try it out during today and report back.

davidmhewitt:
I've just pushed another commit to #663 that might help narrow down the root cause of this too.

If possible, could you test both?

casasfernando:
Sure. I can test both. :)

casasfernando:
> I've just pushed another commit to #663 that might help narrow down the root cause of this too.
> 
> If possible, could you test both?

@davidmhewitt I just updated #650 with a new trace after pulling the latest commits from your branch.

Thanks

casasfernando:
Congrats @dbhowell your patch fixed the crash on startup for me as well!
I will test it a bit more but so far so good.

Thanks.

src/ConversationList/ConversationItemModel.vala
```diff
@@ -153,7 +155,7 @@ public class Mail.ConversationItemModel : GLib.Object {
 
     public int64 timestamp {
```
tintou, line 156:
```suggestion
    public int64 timestamp { get; private set; }
```
and then simply use `timestamp = get_newest_timestamp (new_node, -1);`

tintou:
It looks great, just adding a comment to have a more concise code but it's indeed avoiding issues

tintou (approved):

Issue #659, opened 2021-08-22
Unread email badge next to email folders are only refreshed on app restart


casasfernando:
## What Happened
Unread email badges next to email folders are only updated on app start.
If I read the unread emails in a folder the badge of unread emails is not updated.
If I close and start the app again, then the badge of unread emails is updated with the right number of unread emails.

## Expected Behavior
Unread email badge next to email folders should be updated as soon as the emails in that folder are being read.

## Steps to Reproduce

1. Go to a folder in your email account with unread emails.
2. Read the or mark read the unread emails. This should update the unread email badge intermediately.

### Logs
N/A

## Platform Information
elementary OS 6 Odin
Mail version: io.elementary.mail/focal,now 6.1.0+r613+pkg23~ubuntu6.1

marbetschar:
@casasfernando for now I'm fine with having badge-related stuff in one place, because they don't behave as expected at all - and having all aspects in one place makes it easier to get an overview of what needs to be done. Once they are "supposed to work" we can always make more granular issues.

PR #665, opened 2021-08-29
Reorder conversations

dbhowell:
If a new message arrives, the thread should be at the 'top' of the list and be seen as 'unread'. (see #659 )

Should a thread be "unread" if a single message in the thread has been unread? This is how I understand it as. So when the 5 second delay is up, set all messages in the thread as 'seen' to remove the unread status.


marbetschar:
Hi @dbhowell :wave:,

Thanks for your Pull Request! How can I verify this works from a functional perspective? I tried with multiple mail accounts and replied back and forth - but I don't seem to be able to get any notable difference in behaviour :thinking:

dbhowell:
Currently, if a new email arrives on an older existing conversation, the conservation does not go to the top, nor does it highlight as unread, see screenshots:

file:///home/davidh/Pictures/screen02.png

Note: if you click another folder and then return, the conversation is now at the top but still not higlighted as unread.

src/ConversationList/ConversationListBox.vala
```diff
@@ -107,6 +108,16 @@ public class Mail.ConversationListBox : VirtualizingListBox {
         });
     }
 
+    private static void set_thread_seen (Camel.FolderThreadNode? node) {
```
marbetschar, line 111: Can we rename this function to set_thread_flag and pass the flag as argument?

```vala
private static void set_thread_flag (Camel.FolderThreadNode? node, Camel.MessageFlag flag)
```

marbetschar (changes requested):
Nice job! I can confirm this works as expected 👍 
Just left some nitpicking comments abount method naming and making them a tad more re-usable.

src/ConversationList/ConversationItemModel.vala
```diff
@@ -190,4 +189,14 @@ public class Mail.ConversationItemModel : GLib.Object {
 
         return time;
     }
+
+    private static bool get_all_unread (Camel.FolderThreadNode? node) {
```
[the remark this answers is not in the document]
marbetschar: PS: We should also return as soon as possible to avoid looping over all child nodes. So if the flag was already found:

```vala
if (!has_flag) {
    for (unowned ...) {
        has_flag = ...
        if (has_flag) {
            break;
```

src/ConversationList/ConversationItemModel.vala
```diff
@@ -190,4 +189,19 @@ public class Mail.ConversationItemModel : GLib.Object {
 
         return time;
     }
+
+    private static bool has_thread_flag (Camel.FolderThreadNode? node, Camel.MessageFlags flag) {
+        var has_flag = !(flag in (int)node.message.flags);
+
+        if (!has_flag) {
+        for (unowned Camel.FolderThreadNode? child = node.child; child != null; child = child.next) {
```
marbetschar, line 197:
```suggestion
            for (unowned Camel.FolderThreadNode? child = node.child; child != null; child = child.next) {
```

src/ConversationList/ConversationListBox.vala
```diff
@@ -107,6 +108,16 @@ public class Mail.ConversationListBox : VirtualizingListBox {
         });
     }
 
+    private static void set_thread_flag (Camel.FolderThreadNode? node, Camel.MessageFlags flag) {
+        if (!(flag in (int)node.message.flags)) {
+            node.message.set_flags (Camel.MessageFlags.SEEN, ~0);
```
marbetschar, line 113:
```suggestion
            node.message.set_flags (flag, ~0);
```

marbetschar (changes requested):
Thanks for keeping working on this! 💪
Just two tiny little things, then we're good to go! 🥳

src/ConversationList/ConversationItemModel.vala
```diff
@@ -190,4 +189,19 @@ public class Mail.ConversationItemModel : GLib.Object {
 
         return time;
     }
+
+    private static bool has_thread_flag (Camel.FolderThreadNode? node, Camel.MessageFlags flag) {
+        var has_flag = !(flag in (int)node.message.flags);
```
marbetschar, line 194:
```suggestion
        if (node == null) {
                return false;
        }
        var has_flag = !(flag in (int)node.message.flags);
```

src/ConversationList/ConversationListBox.vala
```diff
@@ -107,6 +108,16 @@ public class Mail.ConversationListBox : VirtualizingListBox {
         });
     }
 
+    private static void set_thread_flag (Camel.FolderThreadNode? node, Camel.MessageFlags flag) {
+        if (!(flag in (int)node.message.flags)) {
```
marbetschar, line 112:
```suggestion
        if (node ==null) {
                return;
        }
        if (!(flag in (int)node.message.flags)) {
```

marbetschar (approved):
Sorry, my bad - I forgot the null checks in my last review. Anyway, I'm approving this now, so feel free to merge once you added them!

Thanks for this PR!

marbetschar (approved):
:partying_face:

PR #667, opened 2021-08-30
vapi: Use Camel vapi from latest master

tintou:
Keep it updated with the latest binding fixes upstream.

### DIFF
--- a/src/Backend/MoveHandler.vala
+++ b/src/Backend/MoveHandler.vala
@@ -30,13 +30,13 @@ public class Mail.MoveHandler {
     private uint timeout_id = 0;
     private MoveType move_type;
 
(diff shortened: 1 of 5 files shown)


marbetschar (approved):
Looks good & works as expected. Just one thing: What's the difference between `unowned` and `weak`  - and when to use either one of those?

I mean, I get why you changed it at some parts: Because this way you don't have to cast to "unowned Type". But I don't think I understand the conceptual difference.

Anyway; feel free to merge!

tintou:
There is currently no difference in Vala, even if technically a weak GObject would mean the use of [GWeakRef](https://docs.gtk.org/gobject/struct.WeakRef.html) but that's not the case currently

Issue #671, opened 2021-08-31
Reply To: Sender account wrong if recipient is not identical


marbetschar:
## What Happened

While https://github.com/elementary/mail/pull/634 introduced a fix for pre-selecting the correct account when replying to an e-Mail, there is still room for improvement: It does not work correctly in cases where the recipient address does not match 1:1 with any configured mailbox account.

Some examples:

- The following recipient addresses are all delivered to the same Gmail inbox:
    - xyz@gmail.com
    - xyz+abc@gmail.com
    - xyz+def@gmail.com
    - xyz+...@gmail.com
- There are other providers, allowing similar things
- Also, if the user configured a CATCH ALL address on his `custom-domain.tld`, it is also possible to forward `anything@custom-domain.tld` to a specific inbox like `my-inbox@custom-domain.tld`

## Expected Behavior

When I reply to a mail which was sent to `xyz+test@gmail.com`, I'd expect the account to send the reply from which is selected to be `xyz@gmail.com`. Similarly, if the mail was sent to `test-abc@custom-domain.tld`, I'd expect `my-inbox@custom-domain.tld` to be preselected.

## Proposed Solution

If there is no exact match, we should test for the mail address domain `@domain.tld`. If there is more than one match with the same domain, we should check if the local part of the address `xyz+abc` contains the prefix of a known mail account's local part within the same domain.

4jNsY6fCVqZv:
> Similarly, if the mail was sent to test-abc@custom-domain.tld, I'd expect my-inbox@custom-domain.tld to be preselected.

If the email was sent to `test-abc@custom-domain.tld`, I would expect a reply to automatically pre-select `test-abc@custom-domain.tld` as the sender.

marbetschar:
@4jNsY6fCVqZv sure 😆 - but only in case this mail account exists. In the scenario described above, the original recipient of the mail was `test-abc@custom-domain.tld` - but this mail box does not exists, therefore the CATCH ALL rule on the server forwarded it to the `my-inbox@custom-domain.tld` account. And the `my-inbox@custom-domain.tld` is also the account configured in elementary Mail.

marbetschar:
Looked at it again and now it behaves correctly since the code in place uses the account mail address for comparison. I probably selected another From address during scrolling, because the conversation is rendered above - but that's one for another issue.

PR #672, opened 2021-08-31
MessageListBox: Hide other children while InlineComposer is shown

marbetschar:
This PR is related to https://github.com/elementary/mail/issues/671 where I accidentally changed the sender address while scrolling.

We fix the weird scrolling behavior here by hiding other messages while the InlineComposer is shown.

**Before**

https://user-images.githubusercontent.com/392542/131533890-aa445f52-f175-4a26-96a6-44fb629bf919.mp4

**After**

https://user-images.githubusercontent.com/392542/131533915-62b68fd1-d44d-43ed-9751-0c889bfb8312.mp4

### DIFF
--- a/src/MessageList/MessageListBox.vala
+++ b/src/MessageList/MessageListBox.vala
@@ -98,10 +98,16 @@ public class Mail.MessageListBox : Gtk.ListBox {
     }
 
     public async void add_inline_composer (ComposerWidget.Type type, MessageListItem? message_item = null) {
+        var children = get_children ();
+
         /* Can't open a new composer if thread is empty or currently has a composer open */
-        var last_child = get_row_at_index ((int) get_children ().length () - 1);
+        var last_child = get_row_at_index ((int) children.length () - 1);
         if (last_child == null || last_child is InlineComposer) {
             return;
+        }
+
+        foreach (var child in children) {
+            child.hide ();
         }
 
         if (message_item == null) {
@@ -121,6 +127,10 @@ public class Mail.MessageListBox : Gtk.ListBox {
             can_move_thread = true;
             remove (composer);
             composer.destroy ();
+
+            foreach (var child in get_children ()) {
+                child.show ();
+            }
         });
         add (composer);
         can_reply = false;


tomaswarynyca (approved):

tintou:
I think that in term of UX I'd prefer to keep the other children as I often need to look again at the conversation, a button to detach the composer into a window would be better I think

marbetschar:
@tintou aren't you seeing the whole conversation within the composer itself? At least on my end it cites the whole conversation - so I can simply scroll down to see older messages within the composer.

UX wise I'm not 100% happy with the way things are implemented right now, because I'm regularly struggling with scrolling into the correct direction as described here: https://github.com/elementary/mail/discussions/609

danrabbit:
Maybe @cassidyjames has thoughts, but personally I don't think it would be a terribly experience if we got rid of the inline composer since it adds quite some complexity

cassidyjames:
Yeah I'm not terribly torn either way. Popping into a new window would let us keep the conversation in the main window, which could be handy for reference if the composer is not modal. But I'm also not really a "heavy" email user. :)

marbetschar:
Personally I like the inline composer - IMHO its way less distracting than popping out a new window. But I agree, the proposed solution here is more like fixing a symptom of a more fundamental problem (details here https://github.com/elementary/mail/discussions/609).

So its probably better to inverse the sort of the whole listbox and add the inline composer at the top - instead of adding the compoesr at the bottom and hide messages while its visible?

danrabbit:
I think when we looked at other mail clients, the list of messages was almost always with the newest messages on the bottom and reversing the direction would break the collapsing behavior we currently have

I also don't think hiding messages while composing is a good idea because, as Cass said, it's easier to reference. Even though messages are quoted, it's still a bit harder to follow than just glancing up to the previous message

danrabbit:
Running a super-duper scientific twitter poll to see if we can get some information about how other people feel: https://twitter.com/DanielFore/status/1441090872007806980

danirabbit:
Closing since we decided to remove the inline composer

PR #674, opened 2021-08-31
Re-adds "Send by Email" contract

### LINKED ISSUE #666: Send via email contractor option
## Problem

When I want to send something via email, I have to open the app and then attach the file to send it.

## Proposal

The old mail app on elementary OS 5 had used contractor to provide a send via email option when right clicking on a file.

## Prior Art

elementary OS 5 mail app

marbetschar:
Fixes #666.

### DIFF
--- a/data/io.elementary.mail.attach
+++ b/data/io.elementary.mail.attach
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# This is a wrapper script to attach several files to an email in Mail.
+# Written by Viko Adi Rahmawan <vikoadi@gmail.com>
+# License: 3-clause BSD
+
+#TODO: define a Gettext domain
+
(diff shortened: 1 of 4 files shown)


Marukesu:
I believe we should make files use the [Email Portal](https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-org.freedesktop.portal.Email) instead of re-adding the contract in mail. Contracts cannot be provided by sandboxed applications, so they will be useless when mail gets flatpaked.

marbetschar:
Makes sense, thanks for pointing that out! Will look into it.

marbetschar:
Converting to draft, since using the Portal is the better solution. Will close this as soon as the feature itself is re-added anyhow.

marbetschar:
Superseded by https://github.com/elementary/mail/pull/682

Issue #646, opened 2021-08-15
Searching mails causes slowdowns


atirut-w:
## What Happened

Searching the mail list cause huge slowdowns when typing

## Expected Behavior

Searching should not slow the app down

## Steps to Reproduce

1. Set up your email account(GMail in my case)
2. Open the Mail app
3. Search

## Platform Information

![Screenshot from 2021-08-14 23-59-17](https://user-images.githubusercontent.com/25323231/129462954-97b2879d-4070-4368-a342-c2a627aba2bc.png)

Issue #287, opened 2018-10-22
Show offline messages upon opening


davidmhewitt:
When opening the rewrite of mail, messages are not displayed in the list until the backend has checked for new messages and downloaded them. I feel as though it's better to show the messages we already have offline while that is happening in the background as it sometimes takes ~10 seconds to download new messages while you're sat looking at a blank window.

Evolution is able to do this.

danrabbit:
For clarity, we can now already load messages instantly offline, but when we are online it takes extra time to startup. So we should load the offline pool first whether we're online or not

PR #676, opened 2021-09-01
Improve search performance

marbetschar:
Attempt to fix https://github.com/elementary/mail/issues/646:

I found the changes proposed here to provide a much better experience of the overall perception of the search performance. This PR also fixes an issue where clearing the search term left us with an empty conversation list - because the search term "" does not return any results.

As a side effect it also seems the startup time of Mail improved by removing the folder synchronization call in `ConversationListBox.load_folder`. However, I'm not quite sure if this also has some unwanted side effects...

The line in question which this PR removes is: `yield folder.refresh_info (GLib.Priority.DEFAULT, cancellable);`


marbetschar:
Maybe related to https://github.com/elementary/mail/issues/287

tintou (approved):

Issue #10, opened 2017-04-25
Show only unread messages


luispablo:
Currently one can mark an email as important, but then I cannot filter such mails from the rest. It would also be usefull if one could filter all unread mails.

Issue #236, opened 2018-08-21
Starred emails folder doesn't show in sidebar


SamMakesCode:
Would be nice to be able to toggle on/off only starred messages. Seems like a fairly big feature that's missing (unless I'm missing something).

danrabbit:
You can see your starred emails by clicking the "Starred" folder in the sidebar under any account

samlittler:
In that case, there's a bug with the menu on the left as I don't have a starred folder.

danrabbit:
Hm Reopening the issue, but I can't confirm

samlittler:
Hi, yes. I hope this is sufficient evidence?

[https://ibb.co/dyOf2e](https://ibb.co/dyOf2e)

cassidyjames:
@danrabbit is it possible this is some special Gmail handling thing?

danrabbit:
@cassidyjames that's very possible. It might be that GMail creates a real folder for this and this isn't a virtual folder

cassidyjames:
@samlittler that's what I'm getting at—it could be that Dan is seeing the starred folder because it's something that's only in Gmail or GSuide accounts, and you're _not_ seeing it because you're not using one. Historically the codebase has done some special handling with Gmail which is why it's relevant.

Ideally, yeah, we could filter down to just starred messages regardless of email provider. But Gmail might be providing "Starred" as yet another folder on the server, which made it easy for those account types.

DanyGee:
I use 3 gmail accounts with pantheon mail and I also don't get starred folder in the pane.

cassidyjames:
That's definitely strange. I've used a handful of Gmail and GSuite accounts and they all get the "Starred" folder. 

@DanyGee, @samlittler: How did you set up the email accounts originally: in Mail, or in System Settings?

DanyGee:
> @DanyGee, @samlittler: How did you set up the email accounts originally: in Mail, or in System Settings?

I did it directly in mail, as usual.
I get starred folder in Mailspring for example.

marbetschar:
+1 for the ability to filter for starred messages. Probably just an activatable icon in the search field would do.

PR #678, opened 2021-09-01
Filter conversations for "Unread" and/or "Starred"

marbetschar:
This PR fixes https://github.com/elementary/mail/issues/10 and https://github.com/elementary/mail/issues/236 by adding the functionality to filter for unread and/or starred conversations in the currently selected folder (see video below).

https://user-images.githubusercontent.com/392542/131774262-de125427-39ac-4ae2-91ab-8e0efc9c4284.mp4


marbetschar:
Managed to get rid of all s-expression warnings. However, the search still does not work as expected when combined with unread/important flags. @tintou is this a bug in upstream - or what am I missing here?

src/ConversationList/ConversationListBox.vala
```diff
@@ -229,16 +231,31 @@ public class Mail.ConversationListBox : VirtualizingListBox {
...
             }
 
-            var sb = new StringBuilder ();
-            Camel.SExp.encode_string (sb, current_search_query);
-            var encoded_query = sb.str;
+            string[] current_search_expressions = {};
 
-            string search_query = """(match-all (or (header-contains "From" %s)(header-contains "Subject" %s)(body-contains %s)))"""
+            if (current_search_hide_read) {
+                current_search_expressions += """(not "read")""";
```
tintou, line 242: Looking at evolution, it should be:
```suggestion
                current_search_expressions += """(not (system-flag "Seen"))""";
```

src/ConversationList/ConversationListBox.vala
```diff
@@ -229,16 +231,31 @@ public class Mail.ConversationListBox : VirtualizingListBox {
...
+            }
+
+            if (current_search_hide_unstarred) {
+                current_search_expressions += """(not (not "important"))""";
```
tintou, line 246:
```suggestion
                current_search_expressions += """(not (system-flag "Flagged"))""";
```

marbetschar:
Thank you @tintou - this did the trick! It works now as expected on my end. However, this is painfully slow to test and would greatly benefit from the changes introduced in https://github.com/elementary/mail/pull/676.

So to whomever is gonna reviewing this, its probably less painfull to review the other PR first :)

tintou (approved):
Works!

src/HeaderBar.vala
```diff
@@ -47,6 +50,37 @@ public class Mail.HeaderBar : Hdy.HeaderBar {
...
+            description = _("Hides all unstarred emails")
+        };
+
+        var filter_menu_popover_grid = new Gtk.Grid () {
+            column_spacing = 6,
+            margin_bottom = 6,
+            margin_top = 12,
+            orientation = Gtk.Orientation.VERTICAL,
+            row_spacing = 6
+        };
```
danrabbit, line 67: We can really reduce the padding here quite a bit
```suggestion
        var filter_menu_popover_grid = new Gtk.Grid () {
            margin_bottom = 3,
            margin_top = 3,
            orientation = Gtk.Orientation.VERTICAL
        };
```

src/HeaderBar.vala
```diff
@@ -47,6 +50,37 @@ public class Mail.HeaderBar : Hdy.HeaderBar {
...
+
+        filter_menu_popover_grid.add (hide_read_button);
+        filter_menu_popover_grid.add (hide_unstarred_button);
+        filter_menu_popover_grid.show_all ();
+
+        var filter_popover = new Gtk.Popover (null);
+        filter_popover.add (filter_menu_popover_grid);
+
+        filter_button = new Gtk.MenuButton () {
+            can_focus = false,
```
danrabbit, line 77: We don't want to prevent buttons from being able to gain focus. This would make it inaccessible with the keyboard
```suggestion
```

src/HeaderBar.vala
```diff
@@ -171,8 +213,22 @@ public class Mail.HeaderBar : Hdy.HeaderBar {
...
+            }
+
+        } else if (has_accent_class) {
+            style_context.remove_class (Granite.STYLE_CLASS_ACCENT);
+        }
+    }
+
     public void set_paned_positions (int start_position, int end_position, bool start_changed = true) {
-        search_entry.width_request = end_position - start_position + 1;
+        search_entry.width_request = end_position - start_position - filter_button.get_allocated_width () - 6;
```
danrabbit, line 231: I guess this should use the end margin of the filter button so we know where this number comes from
```suggestion
        search_entry.width_request = end_position - start_position - filter_button.get_allocated_width () - filter_button.margin_end;
```

danrabbit (changes requested):
I'm not sure switches make the most sense here since we're choosing which items to include as part of a list and not turning a specific feature on or off.

One way we could do this is to change the switch labels to something like "Hide read messages" and "Hide unstarred messages". That would make them actions instead of objects, and then you wouldn't need the explainer text either

It would probably be nice to create some kind of filter icon. I would bet that there might have already been one made for LibreOffice that we could use

marbetschar:
@danrabbit definitely better - thank you very much for adding the required changes to the PR as well!! IMHO this is good to go, what do you think?

danrabbit (approved):
Let's ship it!

Issue #635, opened 2021-08-10
Not getting new email notifications


matfantinel:
## What Happened
On elementary OS 6.0, I'm not getting notifications for new mail. I setup my Fastmail account via IMAP on System Settings, but I am not getting new mail notifications via the Mail app. I've tried different situations, and did not get a notification on any of them:

* After a clean boot, without opening the Mail app (no notifications)
* After opening the app once, then closing it (no notifications)
* With the app open, but not focused (no notifications)

Not sure if intended, but I don't see any process for io.elementary.mail running in the background when it's closed, like there was on 1.0.

Notifications are enabled for the app on System Settings: 
![image](https://user-images.githubusercontent.com/24247035/128934586-fe389ffd-4531-421e-a6e4-f2a50015dbc1.png)

## Expected Behavior
Mail app should display notifications for new mail.

## Steps to Reproduce

1. Setup an IMAP account via the "Online Accounts" panel in System Settings
2. Receive emails
3. No notifications will be displayed

### Logs
Not sure how to grab logs for this, but these are the Evolution-related processes running on my machine. Not sure if any of them is related to fetching emails or not.
![image](https://user-images.githubusercontent.com/24247035/128935080-5b4cb330-1a9a-493f-bb14-4df53971f581.png)

## Platform Information

* elementary OS 6.0 Odin (RC3 with all updates installed)
* Linux 5.11.0-25 generic
* Dell XPS 9570

RajSolai:
Me too experiencing the same issue with Gmail - IMAP , also there is no option for "Always watch for new email"

rasmus91:
Same here. When you try to go to settings, you are sent to online accounts, where you can't really change anything pertaining to mail.

danrabbit:
Please don't make "me too" comments. You can use the :+1: reaction in the issue report to indicate that the issue affects you as well. Making "me too" comments makes it harder to find comments with helpful information

PR #679, opened 2021-09-04
Notifications

marbetschar:
This PR fixes https://github.com/elementary/mail/issues/635 with a background service which notifies about new mails received.

Stuff which still need to be addressed before this can be merged:

- [x] ~~The `folder_changed` signal is not emitted~~ - this was because the folder returned from `Camel.Store.get_inbox_folder` is not a "real" folder. So we need to get the real one with `Camel.Store.get_folder (inbox_folder.full_name...)`
- [x] Notifications do not show the correct icon (related to https://github.com/elementary/notifications/pull/122)

Execute the following to run and test the current code:

```
ninja -C build install
io.elementary.mail --background
```


marbetschar:
Added the fix from https://github.com/elementary/mail/pull/627 - but now getting a SegmentationFault with the following backtrace:

```
#0  0x00007ffff7ec75f4 in g_str_hash () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007ffff7ec5fbf in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007ffff73175d7 in  () at /lib/x86_64-linux-gnu/libcamel-1.2.so.62
#3  0x00007ffff7317870 in  () at /lib/x86_64-linux-gnu/libcamel-1.2.so.62
#4  0x00007ffff7318c88 in camel_sasl_authtype () at /lib/x86_64-linux-gnu/libcamel-1.2.so.62
#5  0x000055555555a008 in mail_camel_session_real_authenticate_sync
    (base=0x5555555b3a60, service=0x555555628520, mechanism=0x7fffd0019ab0 "PLAIN", cancellable=0x555555a24390, error=0x7fffe5cb5840) at ../daemon/CamelSession.vala:94
#6  0x00007ffff7322875 in camel_session_authenticate_sync () at /lib/x86_64-linux-gnu/libcamel-1.2.so.62
#7  0x00007fffec0ad48a in camel_imapx_server_connect_sync () at /usr/lib/evolution-data-server/camel-providers/libcamelimapx.so
#8  0x00007fffec08c7e8 in  () at /usr/lib/evolution-data-server/camel-providers/libcamelimapx.so
#9  0x00007fffec08cda3 in camel_imapx_conn_manager_connect_sync () at /usr/lib/evolution-data-server/camel-providers/libcamelimapx.so
#10 0x00007fffec0b6465 in  () at /usr/lib/evolution-data-server/camel-providers/libcamelimapx.so
#11 0x00007ffff731ce39 in  () at /lib/x86_64-linux-gnu/libcamel-1.2.so.62
#12 0x00007ffff7476c62 in  () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#13 0x00007ffff7f03374 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#14 0x00007ffff7f02ad1 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#15 0x00007ffff6ed6609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#16 0x00007ffff7085293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

marbetschar:
@davidmhewitt I tried [as you suggested](https://github.com/elementary/notifications/pull/118#issuecomment-841344688) in the [latest commit](https://github.com/elementary/mail/pull/679/commits/513c6eb04c4407b80c452c6012b7ffb44705ac45). But the notification center still refuses to show the correct app icon. What am I missing here?

davidmhewitt:
I can't see where you set the application id of the Daemon but I am reading the code on mobile, so I might have missed it. That should be set to the desktop filename, minus `.desktop`

marbetschar:
@davidmhewitt thanks for the hint! Adding the application id nearly fixed it. Strangely I also had to make sure to set `Type=Application` instead of `Type=application`. After I changed both things, the right icon appeared: https://github.com/elementary/mail/pull/679/commits/a1b0d4a7a73508e12dff13bdbffca2031136ac61

That said, I guess this is ready for review 👍

davidmhewitt:
Thinking about this a bit more however...

How well does this work when you consider Flatpak? I know we're not shipping this as a Flatpak right now, but, does the background portal stuff allow running a separate executable to the main executable, and do all the notifications work properly there?

It kind of feels like the Flatpak model would make this more complicated than just having the main executable send notifications, but I don't know enough to say for sure.

marbetschar:
@davidmhewitt I'm not sure how this is supposed to work as Flatpack tbh. But I've seen @Marukesu's pull request for Calendar which introduces support for Flatpack using the background portal: https://github.com/elementary/calendar/pull/681

Therefore I assumed I can solve it the same way and simply start with the daemon approach because Mail is not Flatpacked atm. So we can ship now and extend/rework later.

Marukesu:
> I'm not sure how this is supposed to work as Flatpack tbh. But I've seen @Marukesu's pull request for Calendar which introduces support for Flatpack using the background portal: elementary/calendar#681

So there should be two `.desktop` files for the daemon, one in the `XDG_DATA_DIRS/applications` and another for the autostart.

For Flatpak, the background portal is responsible for generating the autostart one (what elementary/calendar#681 does), but we still need to export a second .desktop file for the notifications server, unless we go the one binary/application route.

marbetschar:
@Marukesu thanks for the clarification. Now this makes a whole lot more sense. I adjusted this PR so it does generate two desktop files for the daemon as you described.

We can surely refactor once we start packaging Mail for Flatpack.

marbetschar:
I just noticed I do have two entries for Mail in `System Settings > Notifications`. I guess this is due to the fact I'm currently deploying one *.desktop file for the Mail app itself and another one for the Daemon:

Screenshot from 2021-09-07 18-51-42

... so I guess we'd really want to embed the daemon within the Mail app itself, so we end up with one binary as suggested by @Marukesu. From what I understand this would remove the second the notification settings entry but still guarantee we are displaying the correct icon in Notification Center.

I'm not 100% sure on how to implement this, but will look into how its done in AppCenter.

marbetschar:
**UPDATE:** I was able to fix the double-entry issue in System `Settings > Notification` with the one-binary approach described above.

This also means we can simply start the background task now with `io.elementary.mail -s`, same way as we do in AppCenter.

data/meson.build
```diff
@@ -13,6 +13,15 @@ foreach i : icon_sizes
     )
 endforeach
 
+daemon_desktop_config = configuration_data()
+daemon_desktop_config.set('EXEC_PATH', libexecdir+'-daemon')
```
danrabbit, line 17: Is this config line needed? It doesn't appear to me that it's used in the .desktop file

danrabbit (changes requested):
I can confirm this works, but I imagine we don't want to shoot off like 20 notifications at a time :sweat_smile:

Could we count up the messages and senders and if we have more than one new message do something like "10 new messages from 3 senders"?

data/meson.build
marbetschar: Good catch! This reminded me to also remove the `libexecdir` variable from the root `meson.build` file 👍

marbetschar:
@danrabbit made the notification a bit more intelligtent and therefore less intrusive.

src/InboxMonitor.vala
```diff
@@ -0,0 +1,228 @@
...
+                GLib.Application.get_default ().send_notification (unseen_message_info.uid, notification);
+
+            } else if (unseen_message_infos_length > 1) {
+                GLib.Notification notification;
+
+                if (sender_names.length == 1) {
+                    notification = new GLib.Notification (_("%u new messages from %s").printf (unseen_message_infos_length, sender_names.iterator ().next_value ()));
+                } else {
+                    notification = new GLib.Notification (_("%u new messages from %u senders").printf (unseen_message_infos_length, sender_names.length));
+                }
```
ryonakano, line 222: I guess we need to use ngettext for languages that have multiple plural forms, maybe something like this?

```vala
GLib.Notification notification;
string messages_count = ngettext ("%u new message", "%u new messages", unseen_message_infos_length).printf (unseen_message_infos_length);

if (sender_names.length == 1) {
    ///TRANSLATORS: The first %s represents the number of new messages translated in your language, e.g. "2 new messages"
    ///The next %s represents the sender name
    notification = new GLib.Notification (_("%s from %s").printf (messages_count, sender_names.iterator ().next_value ()));
} else {
    ///TRANSLATORS: The first %s represents the number of new messages translated in your language, e.g. "2 new messages"
    ///The next %s represents the number of senders
    notification = new GLib.Notification (ngettext ("%s from %u sender", "%s from %u senders", sender_names.length).printf (messages_count, sender_names.length));
}
```

marbetschar: Thanks - that's really good to know! I struggled exactly with this - but then found we can turn the bug into a feature:

If all the messages are from the same sender, it would be nice to know from whom. So instead of simply displaying "3 messages from 1 sender" we are then able to display "3 messages from Ryo Nakano".

Which I think does provide more value because if I see something like "10 messages from 1 sender" I'm probably interested who this guy is anyway 😆 

What do you think, @danrabbit?

marbetschar: ... unless I'm missing something here? Because of the way the notifications are arranged, I thought we don't need the plural/singular distinction...?

danrabbit: Yeah you still need to use ngettext for languages with multiple plural forms, but I agree that if there's only one sender it makes sense to call out that sender specifically rather than say `1`

src/InboxMonitor.vala
```diff
@@ -0,0 +1,233 @@
...
+                    ///TRANSLATORS: The first %s represents the number of new messages translated in your language, e.g. "2 new messages"
+                    ///The next %s represents the sender name
+                    notification = new GLib.Notification (_("%s from %s").printf (messages_count, sender_names.iterator ().next_value ()));
+                } else {
+                    ///TRANSLATORS: The first %s represents the number of new messages translated in your language, e.g. "2 new messages"
+                    ///The next %s represents the number of senders
+                    notification = new GLib.Notification (ngettext ("%s from %u sender", "%s from %u senders", sender_names.length).printf (messages_count, sender_names.length));
+                }
+
+                GLib.Application.get_default ().send_notification (unseen_message_infos.nth_data (0).uid, notification);
```
danrabbit, line 229: Does the UID here represent the ID of the sender or of the account?

danrabbit (commented):
Awesome that's much better! I just have one question about multiple accounts

Screenshot from 2021-09-09 10 51 23

I'm still getting 2 notifications on startup. I wonder if that's because it's per mail account? That's a good thing imo, but maybe we can call that out more clearly in the notification?

I'm not sure exactly how best to split up the information onto the two lines we have. Maybe @cassidyjames has thoughts? But something like

```
5 new messages
from 3 senders for danielle@elementary.io
```

not sure exactly how to make that work for the individual emails. Maybe it's less necessary because you have the full context of the sender and subject?

src/InboxMonitor.vala
marbetschar: This uid is whether from the sender nor the account: its from the first message received in this batch.

danrabbit: Ah okay. In that case, I imagine we should use the ID of the associated account so that when we get replaces working in the notification center it'll show only one up-to-date notification per account instead of having like "10 new messages" and "9 new messages" from the same account saved

marbetschar:
> I'm still getting 2 notifications on startup. I wonder if that's because it's per mail account? That's a good thing imo, but maybe we can call that out more clearly in the notification?

Exactly: The event is emitted once for each inbox folder - so its run once per mail account. I agree, its probably a good idea to also add the display name of this account to the notification. Most probably this would be a good thing to add in the title. I played a bit with it and ended up with the following, what do you think?

Screenshot from 2021-09-09 21-39-31

src/InboxMonitor.vala
```diff
@@ -0,0 +1,238 @@
...
+                GLib.Application.get_default ().send_notification (unseen_message_info.uid, notification);
+
+            } else if (unseen_message_infos_length > 1) {
+                GLib.Notification notification;
+                string messages_count = ngettext ("%u new message", "%u new messages", unseen_message_infos_length).printf (unseen_message_infos_length);
+
+                if (sender_names.length == 1) {
+                    var sender_name = sender_names.iterator ().next_value ();
+
+                    notification = new GLib.Notification (_("%s - %s").printf (sender_name, inbox_folder.parent_store.display_name));
```
danrabbit, line 222: Would it be a little more obvious to use "to" here? As in "Marco to Work"

```suggestion
                    notification = new GLib.Notification (_("%s to %s").printf (sender_name, inbox_folder.parent_store.display_name));
```

src/InboxMonitor.vala
```diff
@@ -0,0 +1,238 @@
...
+                    ///TRANSLATORS: The %s represents the number of new messages translated in your language, e.g. "2 new messages"
+                    notification.set_body (_("%s").printf (messages_count));
```
danrabbit, line 224: Nothing left to translate here right? :)
```suggestion
                    notification.set_body (messages_count);
```

marbetschar: Ok - wasn't 100% sure if ngettext does everything for us. But this is obviously the case so changed this accordingly.

src/InboxMonitor.vala
marbetschar: Yeah, thought about this too. Ended up with a dash because it separates both things more obviously, but don't have too strong feelings about so I changed to "to"

danrabbit (approved):
Awesome, this is a pretty big feature and it seems to work as expected, so let's get it merged and make sure it's well tested for a release at the end of the month!

Nice work :D

PR #680, opened 2021-09-05
Manual Refresh

marbetschar:
This PR add the ability to manually trigger a refresh and fetch messages from backend:

Screenshot from 2021-09-05 14-16-27

~~However, two Gtk issues remain with the current implementation - wasn't able to figure out what's going on here:~~

- [x] ~~The refresh button seems to be active and the default action whenever Mail is started. How to change this?~~ setting `can_focus = false` fixed this
- [x] ~~The animation of the state change seems to be started quite late, not right on click. How to improve?~~ `yield ..async ...` fixed this - thanks @dbhowell!

And one UX question: I'd love to use `<Control>R` for refresh/reload - but it's already used for Reply. Thoughts?


dbhowell:
Using the async `inbox_folder.refresh_info` would improve the UI responsiveness.

Also, I'm not sure iterating the `session.service_list ()` is right. I think you want to iterate through the `session.get_accounts ()` and then get the service from that instead.

marbetschar:
> I'm not sure iterating the `session.service_list ()` is right. I think you want to iterate through the `session.get_accounts ()` and then get the service from that instead.

Good catch! Currently its quite the same - however, its more consistent and makes sure we are always dealing with the same set of accounts now and in the future. Thanks for the hint!

> Using the async `inbox_folder.refresh_info` would improve the UI responsiveness.

Usually yes, but in this case it does not have the desired effect. No clue why. See the test case in the latest commit: There is a notable lag between clicking and the appearance of the spinner. Maybe there's something going on with an animation timeout or so?

marbetschar:
anyone of @elementary/ux an idea how we can avoid focus of the refresh button on application start?

src/HeaderBar.vala
```diff
@@ -43,6 +45,32 @@ public class Mail.HeaderBar : Hdy.HeaderBar {
             _("Compose new message")
         );
 
+        var refresh_button = new Gtk.Button.from_icon_name ("view-refresh", Gtk.IconSize.LARGE_TOOLBAR) {
+            action_name = MainWindow.ACTION_PREFIX + MainWindow.ACTION_REFRESH,
+            halign = Gtk.Align.START,
+            can_focus = false
```
tintou, line 51: That's probably more this:
```suggestion
            can_default = false
```

marbetschar: Unfortunately this does not have the desired effect on my end. The button is still "activated" upon app start, even if `can_default` is set to false. No clue why 🤷

danrabbit: Yeah, same as the filter we don't want to set `can_focus` to false because then you can't access this button with the keyboard which is bad

danrabbit (approved):
We can probably look at default widget focus in a follow up branch, but I think it makes sense to get this functionality in :)

PR #681, opened 2021-09-06
Session: do not use our own Camel.Sasl subclass

tintou:
Use the right one already provided by Camel


marbetschar (approved):
I can confirm this works - in the notification Daemon as well as in the Mail app. That said, this supersedes https://github.com/elementary/mail/pull/627.

PR #682, opened 2021-09-07
Add Support for the org.freedesktop.portal.Email portal

marbetschar:
The changes in this PR basically re-adds support for attachments in mailto queries. This is needed for https://github.com/elementary/files/pull/1822 to work.

Fixes https://github.com/elementary/mail/issues/666 together with https://github.com/elementary/files/pull/1822. Supersedes https://github.com/elementary/mail/pull/674.

### DIFF
--- a/src/Composer/ComposerWidget.vala
+++ b/src/Composer/ComposerWidget.vala
@@ -366,30 +366,37 @@ public class Mail.ComposerWidget : Gtk.Grid {
         }
 
         if (mailto_query != null) {
-            var result = new Gee.HashMap<string, string> ();
+            from_revealer.reveal_child = can_change_sender = true;
+
+            var result = new Gee.HashMap<string, Gee.ArrayList<string>> ();
             var params = mailto_query.split ("&");
 
             foreach (unowned string param in params) {
                 var terms = param.split ("=");
                 if (terms.length == 2) {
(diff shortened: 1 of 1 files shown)


Marukesu (approved):
I can confirm that this is working, i added a little suggestion about using Gee.MultiMap to simplify a bit.

As a side note, on Flatpak, because we don't have access to host filesystem other than via the FileChooser Portal, we end with a unnamed 0 byte attachement. but it's easily fixed by running  `flatpak override --filesystem=home io.elementary.mail` (or `--filesystem=host`), or by implementing our own backend (or do a PR to the gnome one) that export the files to the Document Portal if the app is a flatpak.

marbetschar:
@Marukesu thanks for your review! Adjusted the code accordingly. One thing to note:

When using `Gee.HashMultiMap` I was unable to compile `result["..."].get (0)` - so I had to use `result["..."].to_array ()[0]`

Marukesu:
> When using `Gee.HashMultiMap` I was unable to compile `result["..."].get (0)` - so I had to use `result["..."].to_array ()[0]`

Oh, i see… Strange that `Gee.Collection` doesn't have a get method. Well, i believe `to_array ()[0]` is fine, since the alternative would be `iterator ().next ().get ()`.

Marukesu (approved):

PR #683, opened 2021-09-07
Show time in message header

marbetschar:
Adds the time in the preferred format of the user to the message header. Also the date format was hard coded, ignoring user preferences. This has also been adjusted in this PR.

Wondering if we want a hardcoded comma between date and time though? :thinking: 

**Before**

Screenshot from 2021-09-07 11-04-07

**After**

Screenshot from 2021-09-07 11-03-33


JoseExposito (approved):
Code LGTM.

> Wondering if we want a comma between date and time though?

Probably not, I might not make sense depending on the locale used. cc @ryonakano to hear his opinion before we merge this.

ryonakano:
> Probably not, it might not make sense depending on the locale used.

Yes, it's strange in Japanese at least. However, if it's natural in English, you should make `datetime_label` translatable:

```vala
var datetime_label = new Gtk.Label (new DateTime.from_unix_utc (relevant_timestamp).format (_("%s, %s").printf (date_format, time_format)));
```

This way translators can provide natural translations for their language depends on whether their language should have a comma in this kind of context.

In this case, we should also leave some comment for translators (so that they can understand what are these "%s" represents and connected using comma).

marbetschar:
@ryonakano added the possibility to translate the string along with a comment for translators. Went with "%date at %time".

PR #686, opened 2021-09-09
Meson: rename daemon and remove config

danrabbit:
Since we're not doing anything fancy, I don't think we need that `config` step

Also, I don't think we need keywords here since autostart files aren't searchable afaik

Simplify the file name while we're here because meson lets us do that and it's convenient finding the right file in the sidebar of Code


marbetschar (approved):
Works as expected - thank you for the simplification!

PR #687, opened 2021-09-09
ComposerWindow: close when pressing esc

### LINKED ISSUE #240: Tapping Esc when drafting a mail doesn't save the draft

danirabbit:
Fixes #240

### DIFF
--- a/src/Composer/ComposerWindow.vala
+++ b/src/Composer/ComposerWindow.vala
@@ -62,5 +62,11 @@ public class Mail.ComposerWindow : Hdy.ApplicationWindow {
         window_position = Gtk.WindowPosition.CENTER_ON_PARENT;
 
         add (content_grid);
+
+        key_release_event.connect ((event_key) => {
+            if (event_key.keyval == Gdk.Key.Escape) {
+                close ();
+            }
+        });
     }
 }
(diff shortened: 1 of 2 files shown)


src/Composer/ComposerWindow.vala
```diff
@@ -62,5 +62,11 @@ public class Mail.ComposerWindow : Hdy.ApplicationWindow {
         window_position = Gtk.WindowPosition.CENTER_ON_PARENT;
 
         add (content_grid);
+
+        key_release_event.connect ((event_key) => {
+            if (event_key.keyval == Gdk.Key.Escape) {
+                close ();
+            }
+        });
```
marbetschar, line 70: Any chance we can move this to `ComposerWidget` and connect it there with the `ACTION_DISCARD` - maybe with an accelerator for `ESC` or something? This should then chain up to `composer_widget.discarded.connect (...)` in `ComposerWindow`.

The reasoning with this is, it would also allow us close the inline composer with `ESC` 🥳

marbetschar (changes requested):
That's a great addition! Really love how those little things are coming together to make the experience even more enjoyable and smoother ❤️

Issue #690, opened 2021-09-12
InboxMonitor throws critical and warning messages when running in --background


dbhowell:
## What Happened
When starting with `io.elementary.mail --bakground`, I get several CRITICAL messages and a WARNING message.

## Expected Behavior
I'd expect no warnings/critical messages.

## Steps to Reproduce

1. Add a single IMAP account
2. Start `io.elementary.mail --background`
3. Note criticals and warnings

(Note: I don't have any other "Online Accounts")

### Logs
```
(io.elementary.mail:586323): camel-local-provider-CRITICAL **: 12:39:17.919: scan_fi: assertion 'path != NULL' failed

(io.elementary.mail:586323): camel-local-provider-CRITICAL **: 12:39:17.919: scan_dirs: assertion 'path != NULL' failed

(io.elementary.mail:586323): GLib-CRITICAL **: 12:39:17.919: g_path_is_absolute: assertion 'file_name != NULL' failed

(io.elementary.mail:586323): GLib-CRITICAL **: 12:39:17.919: g_path_is_absolute: assertion 'file_name != NULL' failed

** (io.elementary.mail:586323): WARNING **: 12:39:17.919: InboxMonitor.vala:130: [On This Computer] Error getting inbox folder: Store root (null) is not an absolute path
```
## Platform Information
* Elemtary OS 6 Odin
* Mail compiled from git

dbhowell:
#691 fixes this

PR #691, opened 2021-09-12
Adjust InboxMonitor start

dbhowell:
* Use `get_accounts ()` instead of `list_sources ()` since `session.start ()` does this anyway.
* Simplify the code slightly due to the above
* Removes WARNING/CRITICAL messages (#690)
* `remove_account` probably needs improving


marbetschar:
Thanks for the PR @dbhowell - this looks pretty good! I just have some thoughts I'd love to discuss before we get this merged, but I really like where this is going :+1: 

### Refactoring `account_removed` signal

As we are going to use the `account_removed` signal in `Mail.Backend.Session` now, I think it makes sense to change it to provide account we want to remove: `public signal void account_removed (Mail.Backend.Account account);`

This probably requires some more refactoring in the mail app to handle this signal - but as you already mentioned, this leads to cleaner code overall. From what I've found we need something like the following changes in Mail.Backend.Session for this to work:

```vala
... 
public signal void account_removed (Mail.Backend.Account account);
...
construct {
    accounts = new Gee.LinkedList<Account> (Account.equal_to);
...
public override void remove_service (Camel.Service service) {
    base.remove_service (service);

    if (service is Camel.OfflineStore) {
        account_removed (accounts.remove_at (accounts.index_of (new Account (service))));
    }
}
```

### Simplify the signal handlers

As you already mentioned, the signal change allows us to simplify the signal handlers code. A quick search shows, this probably includes changes in `GroupedFolderSourceItem`, `SessionSourceItem`, `Session` and `InboxMonitor`.

### Using Lock in InboxMonitor

I'm not 100% sure we really need locks here, but its probably safer to do so. That said, we should use `lock` everywhere the containers in question change - not only at removal of accounts.

src/InboxMonitor.vala
```diff
@@ -91,90 +76,55 @@ public class Mail.InboxMonitor : GLib.Object {
 
                     if (inbox_folder != null) {
                         inbox_folder.changed.connect ((change_info) => {
-                            inbox_folder_changed (source, change_info);
+                            inbox_folder_changed (account, change_info);
                         });
-                        inbox_folders.insert (source, inbox_folder);
+                        inbox_folders.insert (account, inbox_folder);
```
marbetschar, line 81: We should probably lock `inbox_folders` for the `insert`.

src/InboxMonitor.vala
```diff
@@ -91,90 +76,55 @@ public class Mail.InboxMonitor : GLib.Object {
...
-                        } else {
-                            debug ("[%s] Automatically checking inbox for new mail is disabled.", display_name);
-                        }
+
+                        debug ("[%s] Checking inbox for new mail every %u minutes…", folder.display_name, refresh_interval_in_minutes);
+                        var refresh_timeout_id = GLib.Timeout.add_seconds (refresh_interval_in_minutes * 60, () => {
+                            inbox_folder_synchronize_sync.begin (account);
+                            return GLib.Source.CONTINUE;
+                        });
+                        synchronize_timeout_ids.insert (account, refresh_timeout_id);
```
marbetschar, line 90: Can we lock `synchronize_timeout_ids` for the `insert` here?

marbetschar (changes requested):
@dbhowell any chance you can add the mentioned locks? Basically just everywhere the containers `inbox_folders` or `synchronize_timeout_ids` are modified.

As @danrabbit mentioned, we can always refactor in subsequent PR's.

marbetschar (approved):
Thank you!!!

marbetschar:
Created a follow up issue for the signal refactoring here: https://github.com/elementary/mail/issues/696

Issue #700, opened 2021-09-23
folder_changed causes crash


dbhowell:
## What Happened
Occasionally mail will crash when an update to a folder happens on the server and mail is trying to update the conversation list. 

## Expected Behavior
When a folder is changed on the server, mail should reflect those changes without crashing.

## Steps to Reproduce

1. Login to mailbox via an alternative method (e.g. webmail)
2. Start mail app
3. Using alternative method, delete TWO emails from the server
4. mail will crash as it is trying to reference an already deleted item

### Logs
```
#0  0x00007ffff7e646d9 in g_type_check_instance_is_fundamentally_a () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#1  0x00007ffff7e457f0 in g_object_get () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#2  0x00005555555a6493 in mail_conversation_item_model_get_deleted (self=0x555557184a50) at ../src/ConversationList/ConversationItemModel.vala:145
#3  0x00005555555ad134 in _mail_conversation_list_box___lambda24_ (self=0x555555bb0480, obj=0x555557184a50) at ../src/ConversationList/ConversationListBox.vala:52
#4  0x00005555555ad182 in __mail_conversation_list_box___lambda24__mail_conversation_list_store_row_visibility_func (row=0x555557184a50, self=0x555555bb0480) at ConversationListBox.c:4021
#5  0x00005555555b0007 in mail_conversation_list_store_get_item_internal (self=0x555555e57060, index=84, unfiltered=1) at ../src/ConversationList/ConversationListStore.vala:70
#6  0x00005555555afcd1 in mail_conversation_list_store_real_get_item_unfiltered (base=0x555555e57060, index=84) at ../src/ConversationList/ConversationListStore.vala:41
#7  0x00005555555cd595 in virtualizing_list_box_model_get_item_unfiltered (self=0x555555e57060, index=84) at ../src/VirtualizingListBox/VirtualizingListBoxModel.vala:32
#8  0x00005555555cdc47 in virtualizing_list_box_model_get_index_of_unfiltered (self=0x555555e57060, item=0x555556dc68d0) at ../src/VirtualizingListBox/VirtualizingListBoxModel.vala:100
#9  0x00005555555b0260 in mail_conversation_list_store_remove (self=0x555555e57060, data=0x555556dc68d0) at ../src/ConversationList/ConversationListStore.vala:91
[...]

PR #701, opened 2021-09-23
Temporarily disable the filter function while batch removals are made

dbhowell:
* Set filter_func to null before batch update
  * Avoids referencing already removed items during the same update (#700)
  * Re-add filter after update


danrabbit (approved):
I can't reproduce this crash right now, but this seems reasonable and it sounds like both @marbetschar and @davidmhewitt think it looks good too, so let's go :) Thanks!

PR #703, opened 2021-09-23
MainWindow: Add an actionbar to conversationlist

### LINKED ISSUE #705: Inconsistency in refresh button between Web and Mail
After the new update, a refresh button was added. It is a hi color icon which is inconsistent with Web's (Epiphany) symbolic icon.

## Mail
Screenshot from 2021-09-24 17-49-13

## Epiphany
Screenshot from 2021-09-24 17-49-33

danrabbit:
Simplifies a few things for us:

Fixes #705
* Moves less frequently used items out of the main headerbar
* Simplifies the headerbar item spacing again and the searchbar lines up with the panes
* Less passing state around between classes
* Gives us a future place to show when the last refresh happened
* Go back to a single symbolic icon for filter with the accent style class to indicate activation

Screenshot from 2021-09-23 12 51 40


marbetschar:
@danrabbit I really like the actionbar approach a whole lot more than what I did with the header bar, well done!! :partying_face: Code looks good as well.

Please note: I also added a small change so the refresh button now refreshes the currently selected folder - instead of just the INBOX of each account. IMHO that's more what I'd expect from the current design and on top of that seems also more practical :)

Feel free to merge if you agree to my changes as well!

PR #710, opened 2021-10-01
Add conversation list context menu

jhaygood86:
Adds a context menu for the single selected conversation.

If we ever support multi-select, this will require some slight refactoring (mainly, to resolve the "state" of the multiple selected items).

### DIFF
--- a/src/ConversationList/ConversationListBox.vala
+++ b/src/ConversationList/ConversationListBox.vala
@@ -103,7 +103,7 @@ public class Mail.ConversationListBox : VirtualizingListBox {
             }
         });
 
-        button_press_event.connect ((e) => {
+        button_release_event.connect ((e) => {
 
             if (e.button != Gdk.BUTTON_SECONDARY) {
                 return Gdk.EVENT_PROPAGATE;
@@ -118,7 +118,7 @@ public class Mail.ConversationListBox : VirtualizingListBox {
             return create_context_menu (e, (ConversationListItem)row);
         });
 
-        key_press_event.connect ((e) => {
+        key_release_event.connect ((e) => {
 
             if (e.keyval != Gdk.Key.Menu) {
                 return Gdk.EVENT_PROPAGATE;
@@ -457,15 +457,17 @@ public class Mail.ConversationListBox : VirtualizingListBox {
 
         var menu = new Gtk.Menu ();
 
-        var delete_menu_item = new Gtk.MenuItem.with_label (_("Delete"));
-        menu.add (delete_menu_item);
+        var trash_menu_item = new Gtk.MenuItem ();
+        trash_menu_item.add (new Granite.AccelLabel.from_action_name (_("Move To Trash"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_MOVE_TO_TRASH));
+        menu.add (trash_menu_item);
 
-        delete_menu_item.activate.connect (() => {
+        trash_menu_item.activate.connect (() => {
             trash_selected_messages ();
         });
 
         if (!item.unread) {
-            var mark_unread_menu_item = new Gtk.MenuItem.with_label (_("Mark As Unread"));
+            var mark_unread_menu_item = new Gtk.MenuItem ();
+            mark_unread_menu_item.add (new Granite.AccelLabel.from_action_name(_("Mark As Unread"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_MARK_UNREAD));
             menu.add (mark_unread_menu_item);
 
             mark_unread_menu_item.activate.connect (() => {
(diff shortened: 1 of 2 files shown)


src/ConversationList/ConversationListBox.vala
```diff
@@ -102,6 +102,32 @@ public class Mail.ConversationListBox : VirtualizingListBox {
                 conversation_selected (((ConversationItemModel) row).node);
             }
         });
+
+        button_press_event.connect ((e) => {
```
marbetschar, line 106: AFAIK are we using `button_release_event` for such things in other places. This has the benefit of not interfering with drag and drop for example. So I think its a good idea to stick to the `some_key_release` events if there's no explicit reasoning for `some_key_press`.

src/ConversationList/ConversationListBox.vala
```diff
@@ -102,6 +102,32 @@ public class Mail.ConversationListBox : VirtualizingListBox {
...
+            var row = get_row_at_y ((int)e.y);
+
+            if (selected_row_widget != row) {
+                select_row (row);
+            }
+
+            return create_context_menu (e, (ConversationListItem)row);
+        });
+
+        key_press_event.connect ((e) => {
```
marbetschar, line 121: Can we use `key_release_event` here? Same reasoning as above :)

src/ConversationList/ConversationListBox.vala
```diff
@@ -425,4 +451,62 @@ public class Mail.ConversationListBox : VirtualizingListBox {
...
         move_handler.expire_undo ();
     }
+
+    private bool create_context_menu (Gdk.Event e, ConversationListItem row) {
+        var item = (ConversationItemModel)row.model_item;
+
+        var menu = new Gtk.Menu ();
+
+        var delete_menu_item = new Gtk.MenuItem.with_label (_("Delete"));
+        menu.add (delete_menu_item);
```
marbetschar, line 461: It would be great to show the accelerator labels in the context menu as well to make keyboard shortcuts discoverable. Granite's `AccelLabel` comes in handy for this.

See the code of the HeaderBar for an example:
https://github.com/jhaygood86/mail/blob/392b9651b79ab6fd1f4bbc9e7068e88f45162763/src/HeaderBar.vala#L115-L118

src/ConversationList/ConversationListBox.vala
```diff
@@ -425,4 +451,62 @@ public class Mail.ConversationListBox : VirtualizingListBox {
...
+
+        delete_menu_item.activate.connect (() => {
+            trash_selected_messages ();
+        });
+
+        if (!item.unread) {
+            var mark_unread_menu_item = new Gtk.MenuItem.with_label (_("Mark As Unread"));
+            menu.add (mark_unread_menu_item);
```
marbetschar, line 469: Making the keyboard shortcut discoverable here is probably a good idea as well :)

src/ConversationList/ConversationListBox.vala
```diff
@@ -425,4 +451,62 @@ public class Mail.ConversationListBox : VirtualizingListBox {
...
+
+            mark_unread_menu_item.activate.connect (() => {
+                mark_unread_selected_messages ();
+            });
+
+        } else {
+            var mark_read_menu_item = new Gtk.MenuItem.with_label (_("Mark As Read"));
+            menu.add (mark_read_menu_item);
```
marbetschar, line 477: Making the keyboard shortcut discoverable here is probably a good idea as well :)

src/ConversationList/ConversationListBox.vala
```diff
@@ -425,4 +451,62 @@ public class Mail.ConversationListBox : VirtualizingListBox {
...
+
+            mark_read_menu_item.activate.connect (() => {
+                mark_read_selected_messages ();
+            });
+        }
+
+        if (!item.flagged) {
+            var mark_starred_menu_item = new Gtk.MenuItem.with_label (_("Mark As Starred"));
+            menu.add (mark_starred_menu_item);
```
marbetschar, line 486: Making the keyboard shortcut discoverable here is probably a good idea as well :)

src/ConversationList/ConversationListBox.vala
```diff
@@ -425,4 +451,62 @@ public class Mail.ConversationListBox : VirtualizingListBox {
...
+
+            mark_starred_menu_item.activate.connect (() => {
+                mark_star_selected_messages ();
+            });
+        } else {
+            var mark_unstarred_menu_item = new Gtk.MenuItem.with_label (_("Unmark As Starred"));
+            menu.add (mark_unstarred_menu_item);
```
marbetschar, line 493: Making the keyboard shortcut discoverable here is probably a good idea as well :)

src/VirtualizingListBox/VirtualizingListBox.vala
```diff
@@ -769,6 +783,10 @@ public class VirtualizingListBox : Gtk.Container, Gtk.Scrollable {
         return null;
     }
 
+    protected void select_row (VirtualizingListBoxRow row) {
+        select_row_internal (row);
+    }
+
```
marbetschar, line 789: Can't we just adjust the visibility of `select_row_internal`? This way we don't have to add another method which is doing the same thing - or am I missing something here?

marbetschar (changes requested):
@jhaygood86 thank you very much for this PR! I just left some notes on minor things I came across - but the code looks really good overall!

src/ConversationList/ConversationListBox.vala
jhaygood86: This is fixed

marbetschar (approved):
Thank you @jhaygood86!

PR #712, opened 2021-10-05
Add sort_func to the MessageListBox

dbhowell:
Sort the MessageListItem on date as occasionally EDS will not iterate in order when adding.

This may be by design but I couldn't see a use-case that fit.

### DIFF
--- a/src/MessageList/MessageListBox.vala
+++ b/src/MessageList/MessageListBox.vala
@@ -38,6 +38,7 @@ public class Mail.MessageListBox : Gtk.ListBox {
 
         get_style_context ().add_class ("deck");
         set_placeholder (placeholder);
+        set_sort_func (message_sort_function);
     }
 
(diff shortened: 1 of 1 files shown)


src/MessageList/MessageListBox.vala
```diff
@@ -126,4 +127,21 @@ public class Mail.MessageListBox : Gtk.ListBox {
         can_reply = false;
         can_move_thread = true;
     }
+
+    private static int message_sort_function (Gtk.ListBoxRow item1, Gtk.ListBoxRow item2) {
+        MessageListItem message1 = (MessageListItem)item1;
+        MessageListItem message2 = (MessageListItem)item2;
```
tintou, line 133:
```suggestion
        unowned MessageListItem message1 = (MessageListItem)item1;
        unowned MessageListItem message2 = (MessageListItem)item2;
```

PR #722, opened 2021-10-13
Live update unread mail badge for INBOX

marbetschar:
This PR adds live update of the unread mail badge count for INBOX folders.

For whatever reason the `folder_changed` signal is not emitted for all other folders on my end. As I don't see a possibility to update their unread mail badge, I made sure they are hidden.

Partly f*xes https://github.com/elementary/mail/issues/659

### DIFF
--- a/data/io.elementary.mail.appdata.xml.in
+++ b/data/io.elementary.mail.appdata.xml.in
@@ -22,17 +22,19 @@
       <binary>io.elementary.mail</binary>
   </provides>
   <releases>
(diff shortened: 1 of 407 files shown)


danrabbit:
Hm, so from what I can tell this only removes the badges from special folders, not user-created folders. I'm kind of on the fence about this anyways. I'm not sure if having no information is better than having old information

marbetschar:
Alright; lets tackle this on 7 then. IIRC the update event was not fired for other folders which might be a bug in EDS solved in a newer version 🤷‍♂️

Issue #697, opened 2021-09-20
Mail rendering issue


marbetschar:
## What Happened

Some mails are not rendered and the Terminal shows the following error:

```
** (io.elementary.mail:41273): CRITICAL **: 20:55:49.891: mail_web_view_add_internal_resource: assertion 'name != NULL' failed
```

## Expected Behavior

The mail should be rendered w/o any error.

## Steps to Reproduce

1. Try to open a marketing mail from Apple in elementary Mail
2. You will see that the mail is not rendered at all

### Logs

Full backtrace:

```
(gdb) backtrace
#0  0x00007ffff7ede295 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007ffff7edf579 in g_logv () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007ffff7edf743 in g_log () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x0000555555580e08 in mail_web_view_add_internal_resource (self=0x5555560168b0, name=0x0, data=0x55555605ead0) at ../src/WebView.vala:172
#4  0x00005555555c6b83 in mail_message_list_item_handle_inline_mime_co (_data_=0x555555f9ad50) at ../src/MessageList/MessageListItem.vala:591
#5  0x00005555555c662e in mail_message_list_item_handle_inline_mime_ready (source_object=0x7fffcc001b00, _res_=0x7fffc8008620, _user_data_=0x555555f9ad50)
    at ../src/MessageList/MessageListItem.vala:583
#6  0x00007ffff7274f49 in  () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#7  0x00007ffff7274f8d in  () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#8  0x00007ffff7ed804e in g_main_context_dispatch () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#9  0x00007ffff7ed8400 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#10 0x00007ffff7ed84a3 in g_main_context_iteration () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#11 0x00007ffff72a2fe5 in g_application_run () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#12 0x0000555555574622 in _vala_main (args=0x7fffffffdb48, args_length1=1) at ../src/Application.vala:142
[...]

PR #736, opened 2021-11-22
avoid assertion error in `add_internal_resource` when `content_id` is…

itmm:
I just put the whole block into a guarding `if`. The code change is minimal, but due to reindentation all the lines in the `if` are marked as changed :-(

I got the assertion triggered in a couple of emails and this change is an easy way to avoid the critical log but keeping the same functionality.

Probably also fixes the issue #697.

### DIFF
--- a/src/MessageList/MessageListItem.vala
+++ b/src/MessageList/MessageListItem.vala
@@ -578,7 +578,9 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
     }
 
     private async void handle_inline_mime (Camel.MimePart part) {
-        if (part.get_content_id () == null) { return; }
+        if (part.get_content_id () == null) {
+            return;
+        }
 
         var byte_array = new ByteArray ();
         var os = new Camel.StreamMem ();
--- a/src/MessageList/MessageListItem.vala
+++ b/src/MessageList/MessageListItem.vala
@@ -578,19 +578,21 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
     }
 
     private async void handle_inline_mime (Camel.MimePart part) {
-        var byte_array = new ByteArray ();
-        var os = new Camel.StreamMem ();
-        os.set_byte_array (byte_array);
-        try {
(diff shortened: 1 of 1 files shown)


davidmhewitt (commented):
Is it not easier/cleaner to invert your if statement and then return early if it's null?

So, if the thing is null, return. Then leave the rest of the method as is.

itmm:
well, some people would say, that multiple returns from a function are a bad thing :-)

I believe it is a bit cleaner, if you state in your if statement under what conditions the block inside is running.

Besides, we should state the change on reducing the complexity of the resulting source, not to make the diff smaller, right?

itmm:
@davidmhewitt if your question was a "rework required", please say so. I got no reaction on my response for a couple of days :-(. I can switch the code as you requested. But I see it as the inferior solution.

jeremypw:
@itmm I agree that in general multiple exits from a function can be problematic especially if they are hidden in the middle of a long function.  However I would make an exception if the other exit is the first piece of code in the function since it is very obvious. It is common for C code functions to have a `g_return_if_xxx ()` call at the start of a function.

I guess if you feel that strongly about it, it is not blocking if this is not changed; it just makes it harder for reviewers to see what has actually changed.

itmm:
no big deal, but I still prefer the original version.

jeremypw:
@itmm  Thanks for changing that.  The only issue is that elementaryos codestyle does not use single line if statements.  You can read more about elementaryos codestyle here:  https://docs.elementary.io/develop/writing-apps/code-style

Sorry for the nit-picking.

src/MessageList/MessageListItem.vala
```diff
@@ -578,6 +578,8 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
     }
 
     private async void handle_inline_mime (Camel.MimePart part) {
+        if (part.get_content_id () == null) { return; }
```
jeremypw, line 581: Elementary code style:
```suggestion
        if (part.get_content_id () == null) {
            return;
        }
```

jeremypw (approved):
Thanks!

PR #747, opened 2021-12-23
Use subject for window title of new message window

### LINKED ISSUE #745: Use subject for window title of new message window
### Problem

The window title for the composer window is not very useful. It provides the bare minimum of info, but doesn't help to distinguish between (for example) multiple composer windows.

### Proposal

Use the subject field to set the window title. I can think of two forms this could take: either the "New Message" title is replaced entirely by the contents of the subject field, or we could concatenate them like "New Message: Do the Thing" (where "Do the Thing" is the subject).

### Prior Art (Optional)

Outlook's composer window is purely the subject. If the subject does not have any contents, the window is titled "Untitled". If there is a subject, that is the title of the window.

meisenzahl:
Closes #745

https://user-images.githubusercontent.com/10796736/147203938-787ef12d-ee9a-45c8-ba8e-c6e793d93481.mp4

### DIFF
--- a/src/Composer/ComposerWidget.vala
+++ b/src/Composer/ComposerWidget.vala
@@ -21,6 +21,7 @@
 public class Mail.ComposerWidget : Gtk.Grid {
     public signal void discarded ();
     public signal void sent ();
+    public signal void updated_subject (string ?subject);
 
     private const string ACTION_GROUP_PREFIX = "composer";
     private const string ACTION_PREFIX = ACTION_GROUP_PREFIX + ".";
@@ -168,6 +169,9 @@ public class Mail.ComposerWidget : Gtk.Grid {
 
         subject_val = new Gtk.Entry ();
         subject_val.margin_top = 6;
+        subject_val.changed.connect (() => {
+            updated_subject (subject_val.text);
+        });
 
         var size_group = new Gtk.SizeGroup (Gtk.SizeGroupMode.HORIZONTAL);
         size_group.add_widget (from_label);
--- a/src/Composer/ComposerWidget.vala
+++ b/src/Composer/ComposerWidget.vala
@@ -21,7 +21,7 @@
 public class Mail.ComposerWidget : Gtk.Grid {
     public signal void discarded ();
     public signal void sent ();
-    public signal void updated_subject (string ?subject);
+    public signal void updated_subject (string? subject);
 
     private const string ACTION_GROUP_PREFIX = "composer";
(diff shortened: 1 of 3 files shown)


src/Composer/ComposerWidget.vala
```diff
@@ -21,6 +21,7 @@
 public class Mail.ComposerWidget : Gtk.Grid {
     public signal void discarded ();
     public signal void sent ();
+    public signal void updated_subject (string? subject);
```
kalmbach, line 24: Hi, looks good! WDYT about changing this to `subject_changed` instead?

src/Composer/ComposerWindow.vala
```diff
@@ -37,20 +37,25 @@ public class Mail.ComposerWindow : Hdy.ApplicationWindow {
...
+        composer_widget.discarded.connect (() => {
+            close ();
+        });
+        composer_widget.sent.connect (() => {
+            close ();
+        });
+        composer_widget.updated_subject.connect ((subject) => {
+            title = subject ?? _("New Message");
+            title = title.length > 0 ? title : _("New Message");
+            titlebar.title = title;
```
kalmbach, line 56: Pardon my ignorance, I very new regarding `Vala` language, I find it hard to think of a scenario where subject could be `null` after being changed by the user. In any case, perhaps we could DRY off the `_("New Message")`, with something like:
```
    bool empty_subject = subject == null || subject.length == 0;                                                                                                               
    titlebar.title = empty_subject ? _("New Message") : subject;
```

meisenzahl:
@kalmbach thank you for your review. Really helpful comments 😊

kalmbach:
Tested! Works smoothly. Nice work!

src/Composer/ComposerWindow.vala
```diff
@@ -37,20 +37,24 @@ public class Mail.ComposerWindow : Hdy.ApplicationWindow {
...
 
+        composer_widget.discarded.connect (() => {
+            close ();
+        });
+        composer_widget.sent.connect (() => {
+            close ();
+        });
+        composer_widget.subject_changed.connect ((subject) => {
+            bool empty_subject = subject == null || subject.length == 0;
+            titlebar.title = title = empty_subject ? _("New Message") : subject;
```
danrabbit, line 55: This might be a little easier to read as:
```suggestion
            if (subject == null || subject.length == 0) {
                subject = _("New Message");
            }

            titlebar.title = title = subject
```

danrabbit (changes requested):

danrabbit (approved):
Nice job!

Issue #737, opened 2021-12-01
Unified Inbox Not Working With EWS Office365


keithcdodd:
### What Happened?

Elementary OS6 Odin, Pantheon Mail. OAUTH2 Office365 and Gmail accounts. Could not figure out a way to add those accounts in the normal online accounts method given OAUTH2 (throws error saying need to be online), but worked around it by downloading Evolution .deb and adding accounts there for both the gmail and the office365 (office 365 was set up through ews). Note for anyone this might help, doing this with Evolution flatpak does not work, needs to be .deb.

Gmail and office365 accounts now show as non-editable online accounts and appear in the pantheon mail app. However, while the gmail shows in the unified inbox, the office365 does not. I can confirm that the office365 is the issue, as I can also add my personal outlook account through the regular online account means in EOS6 and that too populates the unified inbox. Only the office365 through EWS does not appear.

I do not know if this is specifically a thing just for office365, or for all exchange accounts added through EWS protocol through evolution.

Thanks!

### Steps to Reproduce

1. Need to add email accounts with OAUTH2
2. Download .deb of Evolution
3. Add accounts through Evolution, including Office365 OAUTH2 through EWS
4. Accounts appear in online accounts & pantheon mail
5. click on unified inbox -- office365 inbox is not included
6. Issue persists even when deleting the account and trying again

### Expected Behavior

All inboxes appear within the unified inbox, including the office365 EWS accounts. 

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

marbetschar:
OAuth2 authentication for Online Accounts is still in the works. Meanwhile you can add the Office 365 mailbox using IMAP/SMTP as described here: https://support.microsoft.com/en-us/office/pop-imap-and-smtp-settings-8361e398-8af4-4e97-b147-6c6c4ac95353

Adding OAuth2 support is related to https://github.com/elementary/portals/issues/19

keithcdodd:
Thank you. 

Both of my accounts are connected & working in pantheon mail (by adding them in Evolution), it is simply that office365 account's inbox is not also added to the unified inbox, even though the gmail's one is included in the unified inbox.

[...]

PR #753, opened 2022-01-11
Added support for Unified Inbox for Office 365 accounts

marbetschar:
Fixes https://github.com/elementary/mail/issues/737

The root cause is, Office 365 connected through EWS uses `Inbox` as folder name instead of the expected `INBOX`. In the code I mitigated this, by checking for the `backend_name`: If it matches `ews` we are adjusting the folder name accordingly.

If you want to test this, make sure to configure two mail accounts - and one of them connected to Office 356 using the  method described here: https://www.marco.betschart.name/areas/elementary-os/missing-features-and-workarounds#oauth-support-modern-authentication


meisenzahl:
@marbetschar can a general solution be implemented here using [string.ascii_casecmp](https://valadoc.org/glib-2.0/string.ascii_casecmp.html)?

marbetschar:
@meisenzahl indeed. Thanks for the hint!

marbetschar:
@meisenzahl I think _now_ I understand what you meant: Unfortunately this is not possible atm, since the folder name [is passed along `get_folder_info`](https://github.com/elementary/mail/blob/1bd0309f22f1833cc6d3a829de4dd2fc19312216/src/FoldersView/GroupedFolderSourceItem.vala#L98) which in turn fails if the folder name does not match exactly.

So how folder names are compared exactly is up to EDS.

meisenzahl (approved):
LGTM

Issue #734, opened 2021-11-16
Mail hangs when archiving last message in folder


marbetschar:
### What Happened?

Mail app stops responding when archiving the last email in the grouped inbox folder and the below message is displayed. This potentially happens in other folders as well and is probably related to https://github.com/elementary/mail/issues/715

![136166946-9c3a1af8-fe37-4aee-90c1-95276b8bb1ad](https://user-images.githubusercontent.com/392542/141936248-7dd4d13e-d968-4963-922d-01799de60f7b.png)

### Steps to Reproduce

1. Navigate to the grouped inbox folder
2. Start archiving them one by one
3. When you only have one email left in the folder, archive it and Mail will stop responding.

### Expected Behavior

Mail app archives the last email in the folder and continue to work as usual instead of hanging forever.

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

PR #755, opened 2022-01-12
Fix UI freeze because of uint underflow

marbetschar:
This fixes https://github.com/elementary/mail/issues/734 caused by a regression introduced in https://github.com/elementary/mail/pull/717:

Basically when `model.get_n_items () ` returns `0` and we subtract `1` from it, we end up with an underflow because we are dealing with an `uint` here. So the result of this calculation is `uint.MAX`. The following code then starts to loop from `0` to `uint.MAX` multiple times causing my laptop to fly :smile: 

This PR introduces some changes to avoid this.

### DIFF
--- a/data/io.elementary.mail.appdata.xml.in
+++ b/data/io.elementary.mail.appdata.xml.in
@@ -27,6 +27,7 @@
         <p>Improvements:</p>
         <ul>
           <li>Use message subject for compose window title</li>
+          <li>Fix a freeze when archiving the last message in a folder</li>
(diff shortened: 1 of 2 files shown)


tintou:
That's indeed the issue, I would have preferred to replace `var index_max = model.get_n_items () - 1;` by `uint n_items = model.get_n_items ();` and handle the `n_items == 0` case before substracting

marbetschar:
@tintou there you go! :)

tintou (approved):

PR #771, opened 2022-02-06
Show recipient in Sent folder instead of sender

marbetschar:
When we are in the Sent folder, we don't want to show the sender's address in the conversation list because this will always be our own address. Instead, we want to display to whom we've sent the mail.

This PR corrects exactly this behavior.


danrabbit:
@marbetschar can you resolve conflicts here?

tintou (approved):

Issue #462, opened 2020-02-21
Consider multi-paned headerbar design


danrabbit:
![Screenshot from 2020-02-21 09 44 29@2x](https://user-images.githubusercontent.com/7277719/75057930-ea3a4d00-548e-11ea-9167-8429f3e1f12f.png)

danrabbit:
There's some prior art on Dribbble that looks pretty nice: 

![](https://cdn.dribbble.com/users/563897/screenshots/7897478/media/64e9abd78be1135581bd65e65f11e8bc.png)

![](https://cdn.dribbble.com/users/348016/screenshots/3907900/d2.png)

cassidyjames:
Hmmm… at a glance (or at least my initial reaction is that) I'm not a fan; it loses the anchoring effect of the header bar and makes it less clear where the window can be dragged. I think the style can work great for smaller two-pane apps with fewer actions, but in Mail I don't know that it would be better than a standard header bar.

PR #776, opened 2022-04-06
MainWindow: Use multiple headerbars

danrabbit:
Follow up to #462

Gets rid of some resizing voodoo.

Screenshot from 2022-04-06 16 07 04

Clean up the welcome screen:
* Layout more similar to Granite 7's Placeholder
* Use Gtk.Show_uri_on_window for better window management stuff
* Easier window moving with window handle

Screenshot from 2022-05-14 10 28 24

### DIFF
--- a/src/FoldersView/FoldersListView.vala
+++ b/src/FoldersView/FoldersListView.vala
@@ -51,12 +51,14 @@ public class Mail.FoldersListView : Gtk.Grid {
             show_close_button = true
         };
         header_bar.pack_end (compose_button);
+        header_bar.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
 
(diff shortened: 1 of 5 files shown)


hanaral:
@danrabbit The far righthand side of the header bar looks visually unbalanced. An incredibly useful way to utilise this space would probably be backwards and forwards buttons like Files and Web.
I cannot put in to words how nice it would be to have this feature that so many other apps have, and the amount of time everyone would save trying to find 'that one email' across all my mailboxes and folders.

danrabbit:
@hanaral I don't think white space is a bad thing, but regardless this isn't really the place to request new features. If you're having trouble navigating between emails, please open [a new issue report](https://github.com/elementary/mail/issues)

marbetschar (approved):
I love the new look!

PR #780, opened 2022-04-25
Round app icon

danrabbit:
- Match roundness with new mail
- Make sure we're using latest color palette
- clean up while we're here

Todo:
- [x] 16
- [x] 24
- [x] 32
- [x] 48
- [x] 64
- [x] 128


micahilbery (approved):

hanaral:
The new icons don't have the inner contour lines

PR #785, opened 2022-05-23
Translated using Weblate (German)

Soljanka:
native speaker here - I checked all three suggestions on weblate and they are fine


tintou:
Hi there, we are using Weblate for translation changes and are only accepting translations there (here: https://l10n.elementary.io/projects/mail/ )

PR #786, opened 2022-05-24
FoldersListView: move settings here

danirabbit:
Screenshot from 2022-05-23 19 40 28

In line with making sure actions affect the panes they appear in, account settings belongs to the pane that shows accounts.

While we're here push our conversation actions to the start and end of that headerbar instead of using a separator, now that there's nothing on the right side. Conveniently makes space for https://github.com/elementary/mail/pull/783


dmatavel (commented):
Nice changes. It makes me think about the refresh icon. It could be set to the right, side by side with the filter icon -- since the menu/settings icon is on the right side, it would create a pattern. It's just a wander, of course, it's good as it is.

PR #794, opened 2022-07-22
build: support webkit2gtk-4.1

### LINKED ISSUE #793: support for webkit2gtk-4.1 / evolution-data-server 3.45+ / libsoup-3
Starting with GNOME 43, some core components of the desktop have been ported to libsoup-3, which apparently brings a lot of new features, including HTTP/2 support. However, applications and libraries **cannot** load both libsoup-2.4 and libsoup-3 in the same process, as this is unsupported and will crash the application immediately.

The effects of this are especially bad for elementary Mail. It links against evolution-data-server libraries (which have been ported to libsoup-3 as of e-d-s 3.45.1) **and** webkit2gtk-4.0, which is the WebKitGTK library for use with libsoup-2.4.

This means previously built packages of elementary Mail will no longer be installable on Fedora 37 as of today (because it depends on older versions of the e-d-s libraries), and it fails to build, because vala chokes trying to link against both libsoup-2.4.vapi and libsoup-3.vapi.

I asked on the [Fedora development mailing list](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/UA4UYWP73CQSP3LDIMUHTRJPVXHFJJJS/) and apparently the way to resolve this problem is to add support for building against webkit2gtk-4.1 to elementary Mail. I'm not sure how much work that would be, given that Mail's usage of WebKitGTK looks rather more complicated than what other elementary applications do.

bobby285271:
Just packaged evolution-data-server 3.45.1 and built this, seems to work. But I am really not sure what I am doing, @decathorpe can you help test this a bit?

Closes #793


decathorpe:
Have you checked whether this code actually still compiles with older versions of e-d-s? For example, I'm not sure whether replacing all uses of `Soup.URI` with `Glib.Uri` is compatible with both cases, and we might need conditional compilation.

bobby285271:
Ugh... Actually this still builds and runs for me on nixos-unstable (with everything GNOME 42 e.g. e-d-s 3.44.3). For Soup.URI -> GLib.Uri, I actually think that is compatible for both cases, for example gnome-software [is updating that unconditionally](https://github.com/GNOME/gnome-software/commit/661b1c36891f8450046fb8642dda744f34b24439#diff-2d3b0080845918e49f38802098913a9131ff0bf7949f012eee7b4c648e058072) when doing a conditional port :thinking:. I can update that if requested, give me a sec :-)

docs/meson.build
```diff
@@ -30,7 +41,7 @@ all_doc_target = custom_target(
         '@INPUT@'
     ],
     build_by_default: true,
-    input: vala_files,
+    input: [vala_files, config_file],
```
decathorpe, line 44: Why are you adding `config_file` here? This looks like an unrelated change.

bobby285271: I split the change to https://github.com/elementary/mail/pull/795

decathorpe: Awesome, thanks for clarifying and splitting off as a separate change.

decathorpe:
I can confirm that this PR fixes the build against evolution-data-server 3.45+ with webkit2gtk-4.1, and keeps building with earlier versions.

decathorpe:
The patch from this PR no longer applies on top of the master branch, can you please rebase?

webkit-extension/MailPage.vala
```diff
@@ -116,7 +116,11 @@ public class Mail.Page : Object {
 
     private bool on_send_request (WebKit.WebPage page, WebKit.URIRequest request, WebKit.URIResponse? response) {
         bool should_load = false;
+#if HAS_SOUP_3
+        GLib.Uri? uri = GLib.Uri.parse (request.get_uri (), GLib.UriFlags.NONE);
```
meisenzahl, line 120:
```suggestion
        try {
            GLib.Uri? uri = GLib.Uri.parse (request.get_uri (), GLib.UriFlags.NONE);
        } catch (Error e) {
            warning ("Could not parse uri: %s", e.message);
            return should_load;
        }
```

src/Application.vala
```diff
@@ -69,6 +69,26 @@ public class Mail.Application : Gtk.Application {
             string to = null;
 
             try {
+#if HAS_SOUP_3
+                GLib.Uri mailto = GLib.Uri.parse (mailto_uri, GLib.UriFlags.NONE);
```
meisenzahl, line 73:
```suggestion
                try {
                    GLib.Uri mailto = GLib.Uri.parse (mailto_uri, GLib.UriFlags.NONE);
                } catch (Error e) {
                    throw new OptionError.BAD_VALUE ("Argument is not a URL.");
                }

```

meisenzahl (changes requested):
There are 2 warnings about unhandled errors, which I fixed by code suggestion.

meisenzahl (approved):

PR #803, opened 2022-10-01
Docs: Add curiously omitted dependency on meson to the readme

turbobastii:
Every other elementary repo correctly lists `meson` as one of the dependencies to build its application. For some reason this is currently missing from the dependency list here.

This PR includes the dependency on meson as done in every other repo.


lenemter (approved):
Makes sense

PR #824, opened 2023-03-03
flatpak updates

tintou:
Update flatpak and add a small fix

Use rebase to merge it.


io.elementary.mail.json
```diff
@@ -12,7 +12,9 @@
         "--socket=fallback-x11",
         "--socket=wayland",
         "--talk-name=org.gnome.evolution.dataserver.AddressBook10",
+        "--talk-name=org.gnome.evolution.dataserver.Calendar8",
         "--talk-name=org.gnome.evolution.dataserver.Sources5",
+        "--talk-name=org.gnome.evolution.dataserver.UserPrompter0",
```
Marukesu, line 17: I don't think mail uses the Calendar and UserPrompter Services. don't need to expose them to the sandbox.

io.elementary.mail.json
```diff
@@ -50,7 +53,7 @@
                 {
                     "type": "git",
                     "url": "https://gitlab.gnome.org/GNOME/evolution-data-server.git",
-                    "tag": "3.40.2"
+                    "branch": "3.46.4"
```
Marukesu, line 56: I believe this is a tag, not a branch. Also, maybe it's better to keep the version here and in the OS in sync since we aren't really using the server from the sandbox.

io.elementary.mail.json
```diff
@@ -101,10 +89,10 @@
...
                     "sources": [
                         {
-                            "type": "git",
-                            "url": "https://github.com/libical/libical.git",
-                            "tag": "v3.0.9"
-                        }
+                            "type": "archive",
+                            "url": "https://github.com/libical/libical/releases/download/v3.0.16/libical-3.0.16.tar.gz",
+                            "sha256": "b44705dd71ca4538c86fb16248483ab4b48978524fb1da5097bd76aa2e0f0c33"
+                         }
```
Marukesu, line 95: Any reason to prefer the archive over the git repo? if so, wouldn't be better to use archive for all dependencies?

Also, there's a extra space before the close brace.

tintou: Flatpak is way more efficient when dealing with archives than git repositories (it only has to look at if the file is newer on the server before downloading it)

danirabbit:
@tintou is this branch still relevant? If so, can you resolve conflicts

PR #827, opened 2023-03-15
Make SourceList internal

danirabbit:
Since SourceList isn't in Granite 7, we need to make it internal to port to Gtk 4


tintou:
This needs to be ported away from GtkCellRenderer too so I don't think that this is the right way

tintou (approved):
Not a big deal for me

PR #829, opened 2023-03-15
ci: Use JSON instead of YAML

tintou:
Allows to use GNOME Builder directly and makes it easy to copy dependencies from elsewhere.


danirabbit (approved):
Works for me and CI says it's good :+1:

PR #830, opened 2023-03-15
Continue the port to GTK4

tintou:


tintou:
Note that it won't build

danirabbit (approved):

PR #835, opened 2023-03-18
Port to gtk4

### LINKED ISSUE #742: Port Mail to GTK4

leolost2605:
Still very much WIP
Eventually fixes #742 

- [x] Update dependencies
- [x] Comment out what doesn't compile
- [x] Replace unnecessary Grids with Boxes
- [x] Update styles
- [ ] Update Menus according to new HIG
- [x] Reimplement FoldersView using ListView and TreeListModel
    - [ ] Reimplement ability to synchronise with online account after the account was loaded (maybe on a change of the internet connection)
    - [ ] Needs some refinement
- [x] Reimplement ConversationList using ListView (makes VirtualizingListBox obsolete)
    - [x] Reimplement sort and filter functions
    - [x] Scroll to top
- [x] Reimplement MessageList
    - [ ] Reimplement Gravatar Handling
- [x] Update Composer
    - [ ] Handle an error when saving a draft after closing without losing it
- [ ] Update Welcome View
- [x] Update BackendSession authenticate_sync (uses E.CredentialsPrompter which isnt available anymore, any idea where i can find documentation on that or what is a fitting replacement? E.SourceCredentialsProvider?)
    - [ ] make a prompt possible
- [x] Update Dialogs

- [ ] Bug fixing
- [ ] Cleanup
- [ ] Testing

Might have missed something, but will add it when i come across it. Tips, especially on how i could improve some things are welcome and appreciated :)


src/MainWindow.vala
```diff
@@ -180,154 +155,146 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
...
         };
-        filter_menu_popover_grid.add (hide_read_switch);
-        filter_menu_popover_grid.add (hide_unstarred_switch);
-        filter_menu_popover_grid.show_all ();
+       filter_menu_popover_box.append (hide_read_switch);
+       filter_menu_popover_box.append (hide_unstarred_switch);
 
-        var filter_popover = new Gtk.Popover (null);
-        filter_popover.add (filter_menu_popover_grid);
+        var filter_popover = new Gtk.Popover ();
```
danirabbit, line 173: Same link popover, you can remove the manual margins on the box if you add `Granite.STYLE_CLASS_MENU` here

src/MainWindow.vala
```diff
@@ -180,154 +155,146 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
...
+        //placeholder_stack.add_named (welcome_view, "welcome");
 
-        add (placeholder_stack);
+        set_content (placeholder_stack);
 
-        var header_group = new Hdy.HeaderGroup ();
-        header_group.add_header_bar (folders_list_view.header_bar);
-        header_group.add_header_bar (search_header);
-        header_group.add_header_bar (headerbar);
+        //@TODO: lookup a new implementation for Hdy.HeaderGroup
```
danirabbit, line 240: The way to do this in Gtk4 is to use [Gtk.WindowControls](https://valadoc.org/gtk4/Gtk.WindowControls.html) instead and you can set the `side` property

leolost2605:
@danirabbit many thanks for the tips/suggestions, they are much appreciated :)

src/HeaderBar.vala
```diff
@@ -30,95 +30,95 @@ public class Mail.HeaderBar : Hdy.HeaderBar {
...
-        var archive_button = new Gtk.Button.from_icon_name ("mail-archive", Gtk.IconSize.LARGE_TOOLBAR);
+        var archive_button = new Gtk.Button.from_icon_name ("mail-archive");
         archive_button.action_name = MainWindow.ACTION_PREFIX + MainWindow.ACTION_ARCHIVE;
         archive_button.tooltip_markup = Granite.markup_accel_tooltip (
             application_instance.get_accels_for_action (archive_button.action_name),
             _("Move conversations to archive")
         );
 
-        var trash_button = new Gtk.Button.from_icon_name ("edit-delete", Gtk.IconSize.LARGE_TOOLBAR);
+        var trash_button = new Gtk.Button.from_icon_name ("edit-delete");
```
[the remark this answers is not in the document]
leolost2605: STYLE_CLASS_LARGE_ICONS seems to be not available for me. Does it require a certain version of granite and if so which one, as i wasn't able to find it in recent granite release notes? Or am I missing something? And also is there a replacement for STYLE_CLASS_TOOLBAR and STYLE_CLASS_INLINE_TOOLBAR?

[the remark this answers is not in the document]
danirabbit: Looks like it was added in `7.1.0`

There isn't. Toolbar isn't a thing anymore in Gtk4. If you're using `INLINE_TOOLBAR` with Gtk.ActionBar you should use `STYLE_CLASS_FLAT`. But context would help to make sure :)

danirabbit:
Since this has become such a massive diff, I think it would be a good idea to try to break up some of these changes and merge them as we go where possible. Some things like replacing grids with boxes or fixing widget subclasses etc could be done in Gtk 3

src/Composer/ComposerWidget.vala
```diff
@@ -490,18 +498,12 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
-#else
-            var hover_url = url != null ? Soup.URI.decode (url) : null;
-#endif
 
             if (hover_url == null) {
                 message_url_overlay.hide ();
             } else {
                 message_url_overlay.label = hover_url;
-                message_url_overlay.no_show_all = false;
-                message_url_overlay.show_all ();
```
danirabbit, line 504: I think this needs to be replaced with `visible = true`

src/Composer/ComposerWidget.vala
```diff
@@ -262,49 +262,49 @@ public class Mail.ComposerWidget : Gtk.Grid {
...
         action_bar.pack_start (discard);
         action_bar.pack_start (attach);
         action_bar.pack_end (send);
 
         var view_overlay = new Gtk.Overlay ();
-        view_overlay.add (web_view);
-        message_url_overlay = new Granite.Widgets.OverlayBar (view_overlay);
-        message_url_overlay.no_show_all = true;
+        view_overlay.set_child (web_view);
+        message_url_overlay = new Granite.OverlayBar (view_overlay);
```
danirabbit, line 299: Probably needs visible = false
```suggestion
        message_url_overlay = new Granite.OverlayBar (view_overlay) {
            visible = false
        };
```

PR #837, opened 2023-03-27
Application: use startup

danirabbit:
* Gresources with the path rdnn/icons are automatically included in the icon theme
* Use application.startup () so that we don't repeatedly add style providers etc on activate


leolost2605 (approved):
Looks good to me

Btw: Are there any special guidelines for reviewing as I've never done this before? 😳

PR #843, opened 2023-03-29
Replace unnecessary grids with boxes

leolost2605:
This is being done in preparation for a GTK4 port, because a convenience method to add items to grids which have only a single column or row was removed in GTK4. However those grids can be easily replaced with boxes.


danirabbit (approved):
TIL that you don't need to namespace `Gtk.Orientation` :exploding_head: Nice work!

PR #845, opened 2023-03-30
Subclass Gtk.Box instead of Gtk.ListBox in Mail.MessageListBox

leolost2605:
This is being done in preparation for a gtk4 port, as `Gtk.ListBox` will be final in gtk4. Also move the `HeaderBar` from its own class here.

- [x] Change class
- [x] Move its scrolled window to `MessageListBox` from `MainWindow`
- [x] Move headerbar
- [x] Rename to MessageList


leolost2605:
Sorry, wanted to request review from the team, how do i do that? 😳

lenemter:
@leolost2605 Only organization members can request review from the team

src/MessageList/MessageListItem.vala
lenemter, line 1: Why `parent_message_list_box` was introduced here? Is accessing parent somehow broken in GTK 4 version?

src/MessageList/MessageListItem.vala
```diff
@@ -354,8 +357,7 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
     }
 
     private void add_inline_composer (ComposerWidget.Type composer_type) {
-        var message_list_box = (MessageListBox) get_parent ();
-        message_list_box.add_inline_composer.begin (composer_type, this);
+        parent_message_list_box.add_inline_composer.begin (composer_type, this);
```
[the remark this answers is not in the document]
leolost2605: Oh nice, didn't know about that, thank you!

danirabbit:
@leolost2605 regarding the question in the PR description, I agree that moving the headerbar here and renaming this class to "MessageList" makes sense :+1:

src/MessageList/MessageListBox.vala
```diff
@@ -18,12 +18,18 @@
...
     public GenericArray<string> uids { get; private set; default = new GenericArray<string> (); }
 
+    private Gtk.ListBox list_box;
+    private Gtk.ScrolledWindow scrolled_window;
+
     public MessageListBox () {
-        Object (selection_mode: Gtk.SelectionMode.NONE);
+        Object (
+            orientation: Gtk.Orientation.VERTICAL,
+            spacing: 0
```
danirabbit, line 31: I think spacing is 0 by default so I'm not sure you need to set this explicitly

danirabbit (changes requested):
One small inline suggestion. Also don't forget to update po/POTFILES :)

danirabbit (approved):

PR #846, opened 2023-03-31
InlineComposer: Small preparations for GTK4

leolost2605:
Some small preparations for GTK4
- Update margins 
- Construct with properties 
- Update styleclasses


danirabbit (approved):
🚀

PR #848, opened 2023-04-02
Prep for GTK4 MessageListItem cleanup

leolost2605:
- Construct with properties
- Update margins etc.
- Use widget specific API (e.g. attach instead add)


danirabbit (approved):
Looks good! nice cleanup :tada:

PR #849, opened 2023-04-02
Prep GTK4: Include AttachmentBar in MessageListItem

leolost2605:
As `AttachmentBar` is a pretty small class that derives from `Gtk.FlowBox` which is final in GTK4, I think it makes sense to include it in `MessageListItem`.

- Include `AttachmentBar` in `MessageListItem`
- Remove `AttachmentBar`


danirabbit (approved):
Works as expected and makes sense. Simplifies some code too. Nice work!

PR #852, opened 2023-04-07
Proper handling of online/offline states

leolost2605:
Add proper handling of offline/online states:

Fixes an issue that caused mail often to freeze when selecting a message after switching off the internet connection while the app was running.
Fixes an issue that caused mail to sometimes not load folders when offline.


src/MainWindow.vala
```diff
@@ -319,17 +323,27 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
...
                 search_entry.sensitive = false;
             }
         });
 
+        var connection_handler = session.notify["online"].connect (() => {
+            if (session.online) {
+                connection_info_bar.set_revealed (false);
+            } else {
+                connection_info_bar.set_revealed (true);
+            }
```
danirabbit, line 339: could be `connection_info_bar.revealed = !session.online`

src/MessageList/MessageListItem.vala
```diff
@@ -327,7 +327,9 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         expanded = false;
         show_all ();
 
-        avatar.set_loadable_icon (new GravatarIcon (parsed_address, get_style_context ().get_scale ()));
+        if (GLib.NetworkMonitor.get_default ().network_available) {
```
danirabbit, line 330: Good call! Nice catch

src/Backend/Session.vala
```diff
@@ -73,6 +78,30 @@ public class Mail.Backend.Session : Camel.Session {
                 critical (e.message);
             }
         });
+
+        manage_connection.begin ();
+    }
+
+    public async void manage_connection () {
```
danirabbit, line 85: I'm not sure about this. It seems like we'd want to be able to manage the online state per account eventually

src/Backend/Session.vala
```diff
@@ -42,8 +42,13 @@ public class Mail.Backend.Session : Camel.Session {
         Camel.init (E.get_user_data_dir (), false);
         accounts = new Gee.LinkedList<Account> ();
         set_network_monitor (E.NetworkMonitor.get_default ());
-        set_online (true);
+        set_online (false);
         user_alert.connect ((service, type, message) => { warning (message); });
+
+        var network_monitor = GLib.NetworkMonitor.get_default ();
```
danirabbit, line 48: I think we want this to be an unowned reference?
```suggestion
        unowned var network_monitor = GLib.NetworkMonitor.get_default ();
```

src/MainWindow.vala
```diff
@@ -319,17 +323,27 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
...
-        session.account_removed.connect (() => {
+        var account_removed_handler = session.account_removed.connect (() => {
             var accounts_left = session.get_accounts ();
             if (accounts_left.size == 0) {
                 get_action (ACTION_COMPOSE_MESSAGE).set_enabled (false);
                 search_entry.sensitive = false;
             }
         });
 
+        var connection_handler = session.notify["online"].connect (() => {
```
danirabbit, line 334: I think this value could be different per-account so I think we probably need to make sure we're updating this to reflect the online status of the currently selected mailbox. Either that or we should tie this to the network monitor and make sure that the copy reflects that by "offline" we mean disconnected from the network

danirabbit (changes requested):

src/Backend/Session.vala
leolost2605: I think (maybe I am missing something) it's currently not possible without some major development effort to put individual accounts to offline mode, because certain methods (e.g. `Camel.Store.get_folder_info`) automatically try to connect the particular account to the internet if the `Session` is online and I didn't find a way to prevent that (once again I am not really sure, maybe I am missing something). Nevertheless I now moved the method for connecting the individual accounts into the `Mail.Backend.Account` class and added a `local_only` variable to prevent a connection. But for the reason explained above even if set to true this currently doesn't really do anything (So should I remove it?).

src/MainWindow.vala
leolost2605: With the current setup it's basically tied to a `NetworkMonitor` with the advantage of being able to display when the `Session` is manually put to offline mode. I'm not sure, but I think for now this is the most practical way, for the reason explained in the comment above. The only edge case I could think of would be if the server of a certain provider is offline (there might be more so correct me if I'm wrong).

leolost2605: I'll remove the UI part for now, as I am currently working on subclassing a `Gtk.Box` in `ConversationListBox` and will move the corresponding stuff from `MainWindow` to there. I could then implement such a banner per account in a separate PR, what would you think about that? Or should I entirely leave it away for now?

leolost2605:
See #871

PR #853, opened 2023-04-07
MessageList: code style and cleanup

danirabbit:


src/MessageList/MessageList.vala
```diff
@@ -125,20 +107,26 @@ public class Mail.MessageList : Gtk.Box {
...
+        var trash_button = new Gtk.Button.from_icon_name ("edit-delete", Gtk.IconSize.LARGE_TOOLBAR) {
+            action_name = MainWindow.ACTION_PREFIX + MainWindow.ACTION_MOVE_TO_TRASH
+        };
         trash_button.tooltip_markup = Granite.markup_accel_tooltip (
             application_instance.get_accels_for_action (trash_button.action_name),
             _("Move conversations to Trash")
         );
 
+        headerbar = new Hdy.HeaderBar () {
+            show_close_button = true
```
leolost2605, line 127: I think the custom title will be necessary for GTK4 again, when moving back to `Gtk.HeaderBar` so I suppose we could keep it, but I guess it doesn't matter :)

leolost2605 (approved):
Nice, thank you! Wanted to finish it today, but I didn't get around to it.

src/MessageList/MessageList.vala
danirabbit: I think in Gtk 4 we just want to use a WindowControls and not a headerbar so it won't be necessary still

PR #856, opened 2023-04-10
Derive MessageListBox from Gtk.Box in preparation for GTK4

leolost2605:
This is being done in preparation for GTK4

- Derive `MessageListBox` from `Gtk.Box`
- Move corresponding stuff to `MessageListBox` from `MainWindow`
- Cleanup
- Rename to `MessageList`
- Update POTFILES

A few notes:
- The `search_changed` signal of a `Gtk.SearchEntry` has a default delay of 150 ms. I think that's reasonable enough so that we don't have to add another 800 ms of delay. This doesn't really impact performance for small folders, it even makes them seem more reactive and for large folders searching is pretty much broken anyway. Also the delay can be changed in GTK4 if we really would want a longer delay again.
- The `search_entry` and the `filter_button` were set to not be sensitive if no accounts have been added yet. I think this isn't necessary as they can't be focused or activated if there are no accounts present. But maybe I am missing something here, so correct me if I'm wrong.

This also seems to fix all of the annoying `_gtk_widget_get_preferred_size_for_size: assertion 'size >= -1' failed` criticals

### DIFF
--- a/po/POTFILES
+++ b/po/POTFILES
@@ -10,7 +10,7 @@ src/Backend/ContactManager.vala
 src/Composer/ComposerWidget.vala
 src/Composer/ComposerWindow.vala
 src/Composer/InlineComposer.vala
-src/ConversationList/ConversationListBox.vala
+src/ConversationList/ConversationList.vala
 src/ConversationList/ConversationListItem.vala
 src/Dialogs/InsertLinkDialog.vala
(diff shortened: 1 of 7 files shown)


danirabbit (changes requested):
Nice work! Love how this cleans things up :tada:

I only noticed one small issue

PR #857, opened 2023-04-12
MessageListItem: Use Popover instead of Gtk.Menu

### LINKED ISSUE #763: Mail not clearly showing the compose dialog when replying/forwarding from the vertical ellipsis menu option
### What Happened?

When replying/forwarding a message by clicking on the corresponding option in the vertical ellipsis menu on the message view:

image

Mail is not scrolling down to the compose dialog so if the message is long enough, the user will miss the reply dialog unless he scrolls down to the end of the message which may not be obvious. In my case I thought nothing was happening when actually the compose dialog opened down below, "hidden", under the message.

On the other hand if I click on the reply or forward buttons in the top row of Mail app:

image

the message view will automatically scroll down to the compose dialog.

### Steps to Reproduce

1. Open a long message (one that will exceed the message view pane size)
2. Click in the vertical ellipsis menu on the message (under the star)
3. Click on reply, reply all or forward.

### Expected Behavior

Current behavior: Mail is not scrolling down to the compose dialog so if the message is long enough, the user will miss the compose dialog unless he scrolls down to the end of the message which may not be obvious. The compose dialog is actually down below, "hidden", after the message.

Expected behavior: the message view will automatically scroll down to the compose dialog that was opened to reply/forward the message (same behavior as when clicking on the reply or forward buttons on the top)

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Once again preparation for GTK4

Also move `scroll_to_bottom` calls to `MessageList`'s `add_inline_composer` and thus fix #763.

### DIFF
--- a/src/MessageList/MessageListItem.vala
+++ b/src/MessageList/MessageListItem.vala
@@ -203,25 +203,53 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         };
         starred_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
 
-        var reply_item = new Gtk.MenuItem.with_label (_("Reply"));
-        reply_item.activate.connect (() => add_inline_composer (ComposerWidget.Type.REPLY));
+        var actions_menu_box = new Gtk.Box (VERTICAL, 0) {
+            margin_bottom = 3,
+            margin_top = 3
+        };
 
-        var reply_all_item = new Gtk.MenuItem.with_label (_("Reply to All"));
-        reply_all_item.activate.connect (() => add_inline_composer (ComposerWidget.Type.REPLY_ALL));
+        var actions_menu = new Gtk.Popover (null) {
+            child = actions_menu_box
+        };
 
-        var forward_item = new Gtk.MenuItem.with_label (_("Forward"));
-        forward_item.activate.connect (() => add_inline_composer (ComposerWidget.Type.FORWARD));
+        var reply_item = new Gtk.ModelButton () {
+            label = _("Reply")
+        };
+        reply_item.clicked.connect (() => {
+            add_inline_composer (ComposerWidget.Type.REPLY);
+            actions_menu.popdown ();
+        });
 
-        var print_item = new Gtk.MenuItem.with_label (_("Print…"));
-        print_item.activate.connect (on_print);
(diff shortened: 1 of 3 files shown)


src/MessageList/MessageListItem.vala
```diff
@@ -203,33 +203,61 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+            margin_bottom = 3,
+            margin_top = 3
+        };
+
+        var actions_menu = new Gtk.Popover (null) {
+            child = actions_menu_box
+        };
+
+        var reply_item = new Gtk.ModelButton () {
+            label = _("Reply")
```
danirabbit, line 216: if you use `text` instead of `label` these will be left aligned correctly
```suggestion
            text = _("Reply")
```

src/MessageList/MessageListItem.vala
```diff
@@ -203,33 +203,61 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
...
+        };
+        reply_item.clicked.connect (() => {
+            add_inline_composer (ComposerWidget.Type.REPLY);
+            actions_menu.popdown ();
```
danirabbit, line 220: modelbuttons automatically pop down the menu so you don't need this :)

danirabbit (changes requested):
Looks good, just two comments

src/MessageList/MessageListItem.vala
danirabbit: In Gtk4 we'll use `Button` directly and style it with CSS and then yeah it'll be necessary to `popdown` then. So that's fair :+1:

Something I think we should do in a future branch is to maybe make a compose action that accepts the same arguments as `add_inline_composer` so that we can use actions everywhere and not have to have a bunch of different methods. Then we could use GLib.Menu here and save ourselves a lot of trouble :sweat_smile:

leolost2605: Actually I've been working on this and have a somewhat working branch but dropped it as I thought it might get to complex, but I guess I'll start work on that again :)

danirabbit: Cool! Also, I pinged you in Slack with some other thoughts/questions :)

PR #860, opened 2023-04-13
Remove inline composer

leolost2605:
- Remove the inline composer and instead always show a composer window
- Derive ComposerWidget from Gtk.ApplicationWindow and rename to Composer
- Remove other composer implementations 
- move Composer to src from its own folder

Relevant conversation from Slack:

> something I've been thinking about for a while is that I think we should get rid of the inline composer and just always use a separate window
> It would eliminate the issues with like weird nested scrolling stuff, make it easier to implement shortcuts that we don't have to worry about any conflicts, and also it's awkward to try to reference previous emails with the inline composer so could tile the composer window when replying etc

> sounds good to me I think the inline composer really isn't worth the complexity it adds.


danirabbit (approved):
Hell yeah, love the simplification here

PR #862, opened 2023-04-13
AttachmentButton: Updates for gtk4

leolost2605:
- Update margins etc.
- Construct with properties
- Replace individual Gtk.MenuItems with GLib.Menu

### DIFF
--- a/src/MessageList/AttachmentButton.vala
+++ b/src/MessageList/AttachmentButton.vala
@@ -22,6 +22,16 @@ public class AttachmentButton : Gtk.FlowBoxChild {
     public Camel.MimePart mime_part { get; construct; }
     public unowned GLib.Cancellable loading_cancellable { get; construct; }
 
+    private const string ACTION_GROUP_PREFIX = "attachmentbutton";
+    private const string ACTION_PREFIX = ACTION_GROUP_PREFIX + ".";
+    private const string ACTION_OPEN = "open";
+    private const string ACTION_SAVE_AS = "save-as";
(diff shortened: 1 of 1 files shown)


src/MessageList/AttachmentButton.vala
```diff
@@ -22,6 +22,16 @@ public class AttachmentButton : Gtk.FlowBoxChild {
...
 
+    private const string ACTION_GROUP_PREFIX = "attachmentbutton";
+    private const string ACTION_PREFIX = ACTION_GROUP_PREFIX + ".";
+    private const string ACTION_OPEN = "open";
+    private const string ACTION_SAVE_AS = "save-as";
+
+    private const ActionEntry[] ACTION_ENTRIES = {
+        {ACTION_OPEN, on_open},
+        {ACTION_SAVE_AS, on_save_as}
+    };
```
danirabbit, line 33: It would probably be simpler here to just use [`SimpleAction`](https://valadoc.org/gio-2.0/GLib.SimpleAction.html) instead of making an array of action entries. I'm not sure who started this trend of using action entries like this tbh :sweat_smile:

danirabbit (commented):
Looks good, just one comment :)

src/MessageList/AttachmentButton.vala
leolost2605: Like this or is there an even easier way?

danirabbit (approved):
Looks good :tada:

PR #865, opened 2023-04-13
Composer: Remove unused can_change_sender

leolost2605:
- Remove can_change_sender which didn't have any effect
- When opening a draft now the correct sender email address is autoselected


danirabbit (approved):
Looks good! I can confirm this still does the right thing with multiple accounts

PR #866, opened 2023-04-13
Composer: put cc buttons in the headerbar

danirabbit:
Based on #864 

Put the CC buttons in the headerbar to save ourselves a lot of trouble with styling stuff

Screenshot from 2023-04-13 11 39 42


leolost2605 (commented):
Tbh I think the old buttons looked quite a bit better but I don't know how much that's worth to keep.

src/Composer.vala
```diff
@@ -79,12 +79,23 @@ public class Mail.Composer : Hdy.ApplicationWindow {
     }
 
     construct {
+        cc_button = new Gtk.ToggleButton.with_label (_("Cc"));
```
leolost2605, line 82: Also the cc_button is focused and highlighted by default - at least for me - which might be confusing for users so maybe `set_focus (to_val);` seems to be a reasonable default

leolost2605 (approved):
But in the end they're only cosmetic things so it's your call :)

danirabbit:
I'm not super attached to this branch, the other design is probably easier to do in Gtk4 as well :)

PR #867, opened 2023-04-13
MainWindow: Updates for gtk4

leolost2605:
Also in `MoveHandler` `expire_undo` seems to not do anything and therefore was removed along with its timeouts


danirabbit:
Looks like @davidmhewitt added the timeout to avoid a freeze? https://github.com/elementary/mail/pull/333/commits/275747004a5b2e149c134d30db042517bbd0a43f

Maybe he can comment about if that's still needed and we can add a comment here if we do

leolost2605:
Freeze and thaw is now handled within the method like this both when archiving and deleting so I don't think we need it as (also it was already [removed ](https://github.com/elementary/mail/commit/e4ee6721521007779aee7803445c232c948bbe85)from the timeout and the `expire_undo` method) but let's wait for a comment:

```vala
        src_folder.freeze ();

        foreach (var info in moved_messages) {
            info.set_flags (Camel.MessageFlags.DELETED, ~0);
        }

        src_folder.thaw ();
```

danirabbit (approved):
Way simpler, works as expected. Nice job!

PR #869, opened 2023-04-14
Application.css: add scrolledwindow undershoot

danirabbit:
Just a little magic to make it look like headerbars float above scrolled content

Screenshot from 2023-04-13 17 01 05


leolost2605 (approved):
Looks fancy :)

PR #870, opened 2023-04-14
WelcomeView: Updates for gtk4

leolost2605:
- Construct with properties
- Use AppInfo for settings launch


src/WelcomeView.vala
```diff
@@ -72,15 +75,16 @@ public class Mail.WelcomeView : Gtk.Box {
...
+            child = main_box
+        };
 
         add (window_handle);
         show_all ();
 
         welcome_button.clicked.connect (() => {
             try {
-                Gtk.show_uri_on_window ((Gtk.Window) get_toplevel (), "settings://accounts/online", Gdk.CURRENT_TIME);
+                AppInfo.launch_default_for_uri ("settings://accounts/online", null);
```
danirabbit, line 87: We use Gtk.show_uri_on_window here because it includes the event time which fixes issues with focus stealing prevention. `show_uri` still exists in the version of Gtk4 that we have in OS 7, and in future version of Gtk there's `Gtk.UriLauncher.launch`, so I think we should keep this as is for now as it would cause a behavior regression to use AppInfo

danirabbit (changes requested):
Looks good, just one comment

src/WelcomeView.vala
leolost2605: Should then all AppInfo uses be replaced with `show_uri ` in gtk4 (e.g. the one in the menu)? Or is this a special case?

danirabbit (approved):
🚀

PR #871, opened 2023-04-14
Handle network availability

leolost2605:
Add proper handling of network availability:

Fixes an issue that caused mail often to freeze when selecting a message after switching off the internet connection while the app was running.
Fixes an issue that caused mail to sometimes not load folders when offline.

Also see #852

### DIFF
--- a/src/Backend/Account.vala
+++ b/src/Backend/Account.vala
@@ -20,16 +20,23 @@
 
 public class Mail.Backend.Account : GLib.Object {
     public Camel.Service service { get; construct; }
-    public bool local_only { get; set; default = false; }
 
     public Account (Camel.Service service) {
         Object (service: service);
     }
 
+    construct {
+        unowned var network_manager = GLib.NetworkMonitor.get_default ();
+        network_manager.network_changed.connect (() => {
+            manage_connection.begin (network_manager.network_available);
+        });
+        manage_connection.begin (network_manager.network_available);
+    }
+
     public async void manage_connection (bool online) {
         var offlinestore = (Camel.OfflineStore)service;
 
-        if (!local_only && online) {
+        if (online) {
             try {
                 yield offlinestore.set_online (true, GLib.Priority.DEFAULT, null);
                 yield offlinestore.synchronize (false, GLib.Priority.DEFAULT, null);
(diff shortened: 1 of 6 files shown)


src/MessageList/MessageListItem.vala
```diff
@@ -333,7 +333,9 @@ public class Mail.MessageListItem : Gtk.ListBoxRow {
         expanded = false;
         show_all ();
 
-        avatar.set_loadable_icon (new GravatarIcon (parsed_address, get_style_context ().get_scale ()));
+        if (GLib.NetworkMonitor.get_default ().network_available) {
```
danirabbit, line 336: While we're here, should we also listen to changes and try to set this if we get a network connection?

src/Backend/Session.vala
```diff
@@ -41,9 +41,13 @@ public class Mail.Backend.Session : Camel.Session {
     construct {
         Camel.init (E.get_user_data_dir (), false);
         accounts = new Gee.LinkedList<Account> ();
-        set_network_monitor (E.NetworkMonitor.get_default ());
```
danirabbit, line 44: It looks like @tintou added this in. I wonder if we can get a comment from him here?

src/Backend/Account.vala
```diff
@@ -20,9 +20,38 @@
...
+            } catch (Error e) {
+                critical (e.message);
+            }
+            return;
+        }
+
+        try {
+            yield offlinestore.set_online (false, GLib.Priority.DEFAULT, null);
+        } catch (Error e) {
+            critical (e.message);
```
danirabbit, line 51: Should this be a user facing error? If we're unable to come online that seems like something we should throw a dialog for maybe?

leolost2605: Probably should yeah

src/MessageList/MessageListItem.vala
leolost2605: For some reason it doesn't work when connecting directly to the network_changed signal and as gravatar has to be entirely rewritten with gtk4 anyway I'm not sure the benefit and use case is big enough to invest more time in this ¯\_(ツ)_/¯

src/Backend/Session.vala
tintou: It is what Evolution itself is doing https://gitlab.gnome.org/GNOME/evolution/-/blob/master/src/libemail-engine/e-mail-session.c#L1250 unsure why we don't want it here?

leolost2605: I'm not sure either so I added it back :)

danirabbit (approved):
LGTM :rocket:

PR #872, opened 2023-04-14
Composer: set parent internally, use present

danirabbit:
Also fix some places where we assume that `active_window` is a `MainWindow` and not a `Composer`

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -82,10 +82,10 @@ public class Mail.Application : Gtk.Application {
                 to = Soup.URI.decode (mailto.path);
 
                 if (main_window.is_session_started) {
-                    new Composer (main_window, to, mailto.query).show_all ();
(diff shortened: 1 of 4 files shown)


leolost2605 (changes requested):
In `Application` the instantiation of the `Composer` has to be changed in the SOUP 3 part too, I think

leolost2605 (approved):
👍
I only got one (maybe very stupid) question: why do we set `transient_for` on the composer window? What does it actually do in this case?

danirabbit:
@leolost2605 it makes it so the composer window can’t go behind the main window and so that when you move the main window to another workspace, its composer children will move with it.

it might be that this is actually not desirable tbh. I don’t have super strong feelings about it

leolost2605:
Ah ok thank you 😅

I personally think that's quite useful in this case 🤷

PR #874, opened 2023-04-16
WebView: Updates for gtk4

leolost2605:
Use `height_request` instead `override void get_preferred_height`


danirabbit:
Looks like this used to use `height_request` and @davidmhewitt switched it for `preferred_height`. Maybe he has more insight about why that was done? https://github.com/elementary/mail/pull/74/files

PR #877, opened 2023-04-19
Gtk4 port 

leolost2605:
As far as I know most features from the gtk3 version are now implemented. It still has some (mainly styling) issues (see below) and if there is anything that could still be back ported please tell me

GTK3:
image

GTK4:
image

A few general things:
- `SourceList` was replaced with a `ListView` and a `TreeListModel`
- `VirtualizingListBox` was replaced with a `ListView` 

A few issues to note:

- [ ] ReadMe wasn't updated yet (new deps: libadwaita, gtk4, libedataserverui4 >= 3.45.1,  libedataserver >= 3.45.1, granite-7 >= 7.1)
- [ ] Flatpak manifest wasn't updated yet
- [ ] @danirabbit any suggestion on how to style the sidebar? Does it need to be exactly like the sourcelist was? Because currently it's way further indented because of how `Gtk.TreeExpander` handles indentation. This can be changed with GTK 4.10 however. It's probably best to make GTK 4.10 mandatory anyway because of some bugs it fixes with `ListView`. Also I don't know about the account icons I just added them because I thought they might look good :)
- [ ] Some more styling issues, especially with the `Gtk.Paned` separators (at least for me) and the unread badges and their size and alignment (see TODOs). I'm gonna need some help with these :)
- [ ] The menu for the conversations can't be activated with the menu key anymore (needed? I couldn't find an easy way to do this, but it's definitely possible)
- [ ] I'm not sure whether Gravatars work, sometimes it seems so other times it doesn't but at least the UI won't freeze anymore
- [ ] Position isn't remembered anymore. I heard it should be left to the window manager entirely so I don't know 🤷

Feedback and suggestions are always appreciated! 😀


src/Composer.vala
```diff
@@ -931,26 +915,32 @@ public class Mail.Composer : Hdy.ApplicationWindow {
...
-            box.add (image);
-            box.add (name_label);
-            box.add (size_label);
-            box.add (remove_button);
+            box.append (image);
+            box.append (name_label);
+            box.append (size_label);
+            box.append (remove_button);
+
+            margin_top = 3;
```
danirabbit, line 935: Good call moving margin here

src/Dialogs/InsertLinkDialog.vala
```diff
@@ -40,35 +40,34 @@ public class InsertLinkDialog : Granite.Dialog {
...
-        get_content_area ().add (grid);
+        get_content_area ().append (grid);
 
         add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
 
         var insert_button = add_button (_("Insert Link"), Gtk.ResponseType.APPLY);
-        insert_button.can_default = true;
-        insert_button.has_default = true;
+        // insert_button.can_default = true;
+        // insert_button.has_default = true;
```
danirabbit, line 65: This is now [`Gtk.Window.set_default_widget ()`](https://valadoc.org/gtk4/Gtk.Window.set_default_widget.html)

src/MainWindow.vala
```diff
@@ -133,39 +137,44 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
             }
         });
 
-        paned_start = new Gtk.Paned (Gtk.Orientation.HORIZONTAL);
-        paned_start.pack1 (folders_list_view, false, false);
-        paned_start.pack2 (conversation_list, true, false);
+        paned_start = new Gtk.Paned (Gtk.Orientation.HORIZONTAL) {
+            shrink_start_child = false,
+            shrink_end_child = false,
+            wide_handle = true
```
danirabbit, line 143: Is there a reason why you added the `wide_handle` here?

src/MessageList/MessageList.vala
```diff
@@ -33,157 +36,157 @@ public class Mail.MessageList : Gtk.Box {
...
-        add (scrolled_window);
+        append (headerbar);
+        append (scrolled_window);
+
+        var settings = new GLib.Settings ("io.elementary.mail");
+        settings.bind ("always-load-remote-images", load_images_menuitem, "active", SettingsBindFlags.DEFAULT);
+
+        account_settings_menuitem.clicked.connect (() => {
+            try {
+                AppInfo.launch_default_for_uri ("settings://accounts/online", null);
```
danirabbit, line 163: We want to use `Gtk.show_uri ()` here to avoid focus stealing prevention

src/WelcomeView.vala
```diff
@@ -20,10 +20,8 @@
 
 public class Mail.WelcomeView : Gtk.Box {
```
danirabbit, line 21: I think most of this can be replaced with Granite.Placeholder

danirabbit (commented):
Hell yeah, getting a lot closer! Left some comments as I was going through.

Is there any way we could land the new accounts list in the Gtk3 version or are we relying on features in Gtk4?

Unfortunately we're stuck with Gtk 4.6.6 until we rebase on Ubuntu 24.04 so we either have to wait until then to land the Gtk 4 port or stick with what is available in that version

src/MainWindow.vala
```diff
@@ -18,17 +18,14 @@
  * Authored by: Corentin Noël <corentin@elementary.io>
  */
 
-public class Mail.MainWindow : Hdy.ApplicationWindow {
+public class Mail.MainWindow : Adw.ApplicationWindow {
```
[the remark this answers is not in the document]
leolost2605: `Adw.ApplicationWindow` prevents the window manager (or is it gtk that does it? 😅) from adding the standard title bar. I am not sure whether that's the only way to do it but I don't think `set_titlebar` would work because we've got multiple `HeaderBar`s.

src/MainWindow.vala
leolost2605: I hoped this would fix some style issues with the separators because the black border on the left side of the conversation list is shorter than on the right side but it didn't and I'm not even sure whether the black borders belongs to the separators or to something else entirely. Any idea about the origin of these borders and how to fix their size?

src/MainWindow.vala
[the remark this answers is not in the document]
danirabbit: Ah okay yeah, so the way we've been dealing with this in other places is `set_titlebar (new Gtk.Grid () { visible = false });`. I'm not sure if using Adw.ApplicationWindow will have any unintended consequences on styling etc

danirabbit:
Yeah exactly we have to ship Mail in deb currently because of issues with accessing the host keychain with evolution and it not working with online accounts. So we're stuck with Ubuntu's packages there

leolost2605:
I guess we'll have to wait for 24.04 then anyway. In the meantime is there anything I can help with getting the flatpak version to work? I've come over [this](https://github.com/elementary/portals/pull/20#issue-1029959697), is it still relevant?

danirabbit:
I think @tintou probably knows the most about this since he's done upstream work with EDS but yeah my understanding is that the Flatpak EDS can't access the credentials stored in the host EDS

PR #881, opened 2023-04-28
Add move capability

### LINKED ISSUE #657: Move deleted mails to Trash folder
A follow up on #647: The current implementation of deleting a mail is troublesome, because it sets the delete flag of the message and there is way to recover a deleted message once the undo timeout ran down.

A better way of doing this would be not to set the delete flag, but rather move the message to the [Trash folder](https://valadoc.org/camel-1.2/Camel.Store.get_trash_folder.html) and let the server decide when to explicitly delete it.

See the implementation of archiving a message, we can probably refactor this a bit to allow us to move messages either to the Archive or Trash folder. In any case, when we use [`folder.transfer_messages_to`](https://valadoc.org/camel-1.2/Camel.Folder.transfer_messages_to.html) we need to watch out for its returned `transferred_uids` value: When we moved messages to archive folder in the past, this value was always returned as `null` for yet unknown reasons.

### LINKED ISSUE #800: Missing move ability
### What Happened?

Selecting move conversation only allows move to archive. Just switched to OS6 so assume missing code on new rewrite of Mail. This also includes the drop/drag method of moving emails to correct folders. This feature a must due to Yahoos service sends some emails to spam, mostly competitors of yahoo and att. I dislike their censorship but at least I could move to inbox to read as email was composed. I now have several emails in archive which I can not move back out and into correct locations, inbox now archive and inbox. PS can not build pre 6 series pantheon-mail due to ubuntu 18.04 code restrictions of dependencies.

### Steps to Reproduce

1. Select toolbar icon by usual right click
2. Moves to Archive no context menu for location
3. Or try drop/drag.
4. Attempted even old Geary alt/ctrl attempts fail

### Expected Behavior

To be able to place into one of the locations listed in sidebar, or from one location to another.

### OS Version

6.x (Odin)

### Software Version

### Log Output

```shell
No output failure
```

### Hardware Info

Ryzen 2400g, linux latest firmware updates, ubuntu 20.04.3 base, kernel 5.18.12, all software less than week old including developer files.

leolost2605:
Add move capability and unify trash/archive/move actions.

This is somewhat based on #867 as it introduces a whole new system of expiring and undoing.

I'm not sure about the UX part, especially the icon and the menu itself so feedback there is needed.

Fix #800 
Fix #657 but requires elementary/switchboard-plug-onlineaccounts#258 for this


leolost2605:
I'll split up this PR into multiple smaller ones for easier review

danirabbit:
@leolost2605 should this be closed or does it contain something else that wasn't covered in #895?

leolost2605:
@danirabbit no all done, I'm closing it

PR #882, opened 2023-04-29
Use flatpak portal for background and autostart

### LINKED ISSUE #880: Use the background portal
We are currently using the .deb package to install the background service, we should install it at first run via the background service portal https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Background

leolost2605:
Use the flatpak background portal to install the autostart file and request background permission.

Uses libportal.

Requires elementary/portals#73

Closes #880


leolost2605:
Currently maybe blocked by the same as [this](https://github.com/elementary/calendar/issues/778#issuecomment-1523702637)

meisenzahl:
We could prepare shipping Mail as Flatpak by guarding your code with something like this:

```vala
    public bool is_running_flatpaked { 
         get { 
             return FileUtils.test ("/.flatpak-info", FileTest.IS_REGULAR); 
         } 
     }
```

leolost2605:
I updated it so that mail will only use the flatpak portal if it's sandboxed, so in theory this PR is ready for review again. 
However i'm not sure but it seems using the background portal is currently blocked entirely by that the gtk implementation was removed by default with [this release](https://github.com/flatpak/xdg-desktop-portal-gtk/releases/tag/1.10.0) as it depends on gnome-shell. Should I open an issue for this in elementary/portals?

leolost2605:
Another thing currently the background daemon isn't started automatically after it was granted permission on first run, it only starts after restart. Should this be changed?

Marukesu:
> Another thing: currently the background daemon isn't started automatically after it was granted permission on first run, it only starts after restart. Should this be changed?

in the calendar PR, i changed it to always start with the application, the `--background` flags only signalize that no window should be shown.

leolost2605:
Makes sense to me I'll do the same here then :)

Marukesu:
> Also note that currently it's shown as deactivated in switchboard. It still works unless you activate it once and then deactivate it again. That's because of some deprecated stuff that's still used there, I plan on fixing that though.

elementary/switchboard-plug-applications#184 is the fix for that.

src/Application.vala
```diff
@@ -184,6 +191,23 @@ public class Mail.Application : Gtk.Application {
...
+        var portal = new Xdp.Portal ();
+
+        Xdp.Parent? parent = active_window != null ? Xdp.parent_new_gtk (active_window) : null;
+
+        var command = new GenericArray<weak string> ();
+        command.add ("io.elementary.mail");
+        command.add ("--background");
+
+        try {
+            return yield portal.request_background (parent, _("Mail needs to run in background in order to send notifications."), (owned) command, Xdp.BackgroundFlags.AUTOSTART, null);
```
Marukesu, line 205: this line pass the 120 line limit, would be good to break it.

```suggestion
            return yield portal.request_background (
                 active_window != null ? Xdp.parent_new_gtk (active_window) : null,
                _("Mail needs to run in background in order to send notifications."),
                (owned) command,
                Xdp.BackgroundFlags.AUTOSTART,
                null
            );
```

Also, IIRC, the background frontend do not check if the application already has permission, so this would end in the permission dialog begin shown in every launch even when the user already give permission.

in the calendar PR, it stores if it had an successful request and skip the request_background call in the next launches.

Marukesu (changes requested):
Don't make the request_background call in the startup() method (i would say the call to activate here is wrong too), this will mess with DBus activation.

The way to go here is:
  1. call only `InboxMonitor.start()` in the `startup()` method.
  2. in `activate()` check if the background flag got used, if so call `hold()` and then request background, not having a parent window shouldn't be a issue.
  3. if the request fail, call `release()`.

there's no need to make the request if we open an window. instead, we should make the request in the window's delete_event and hold() if permission was granted. but this one can be done in a follow-up.

leolost2605:
> there's no need to make the request if we open an window. instead, we should make the request in the window's delete_event and hold() if permission was granted. but this one can be done in a follow-up.

If that's okay from a design perspective I would integrate this here because it makes implementing the other suggestions quite a bit easier

> Also, IIRC, the background frontend do not check if the application already has permission, so this would end in the permission dialog begin shown in every launch even when the user already give permission.

Actually I think it does [here](https://github.com/flatpak/xdg-desktop-portal/blob/dbe8633d08dd9f790216df3e8104d82fb555c16f/src/background.c#L728). However it would still overwrite any settings made by the switchboard plug every time mail is launched. On the other hand it allows to disable the autostart via flatpak permission which might be the way to go for the future?

leolost2605:
Ok now I implemented it a bit different:
Always `hold ()` and always start a new InboxMonitor in `startup ()`. If the background flag is set just don't open a window. 
Now when closing a MainWindow request background access. If the request fails call `release ()` if it succeeds do nothing. This is based on the assumption that a user doesn't randomly run mail with the background flag so that we can assume that if it is set we already successfully requested background access/installed the autostart file and IMO we then don't necessarily have to check it here again. 
Would doing it like this be ok?

src/MainWindow.vala
```diff
@@ -214,6 +214,16 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
             session_started ();
         });
 
+        delete_event.connect (() => {
+            unowned var application = (Application)GLib.Application.get_default ();
+            application.request_background.begin ((obj, res) => {
+                application.request_background.end (res);
+                destroy ();
+            });
```
Marukesu, line 222: MainWindow already has a application reference, no need to call Application.get_default() and you don't need to call end() if you won't catch any error or work with the return value in async methods.
```suggestion
            application.request_background.begin (() => destroy ());
```

src/Application.vala
```diff
@@ -140,18 +140,13 @@ public class Mail.Application : Gtk.Application {
...
         set_accels_for_action ("app.quit", {"<Control>q"});
 
-        /* Needed to ask the flatpak portal for autostart and background permissions with a parent window
-           and to prevent issues with Session.start being called from the InboxMonitor first */
-        if (!run_in_background) {
-            activate ();
-        }
-
         new InboxMonitor ().start.begin ();
         hold ();
```
Marukesu, line 144: i insist, hold() shouldn't be called in startup(), this will keep the application alive even from a activation without any further action.

leolost2605: Oh sorry I thought you were only referring to `request_background ()`

src/Application.vala
```diff
@@ -190,6 +186,31 @@ public class Mail.Application : Gtk.Application {
...
+                _("Mail needs to autostart and run in background in order to send notifications even while closed."),
+                (owned) command,
+                Xdp.BackgroundFlags.AUTOSTART,
+                null
+            )) {
+                release ();
+            }
+        } catch (Error e) {
+            warning ("Failed to request autostart permissions: %s", e.message);
+            release ();
```
Marukesu, line 210: i think we should let the application go to background if the portal call fail, the worse this can do is causing NotifyBackground to be called in the impl.

src/Application.vala
```diff
@@ -190,6 +186,31 @@ public class Mail.Application : Gtk.Application {
...
+        Xdp.Parent? parent = active_window != null ? Xdp.parent_new_gtk (active_window) : null;
+
+        var command = new GenericArray<weak string> ();
+        command.add ("io.elementary.mail");
+        command.add ("--background");
+
+        try {
+            if (!yield portal.request_background (
+                parent,
+                _("Mail needs to autostart and run in background in order to send notifications even while closed."),
```
danirabbit, line 201: I think we shouldn't be relying on the app to name itself or the permissions its requesting. That should be the responsibility of the portal. So we should only have here the reason like:
```suggestion
                _("Send notifications when new mail arrives."),
```

Marukesu:
> Should I guard the hold with something like a first_activation variable? I saw that in the calendar PR it wasn't so I dont know...

Yes, it will be problematic when sandboxed, ask_background will always be true when running from host, so the amount of hold() done doesn't matter we will always hold. will update that there.

src/Application.vala
danirabbit: It seems like we aren't in charge of that yet, so editing my suggestion

src/Application.vala
leolost2605: For some reason the libportal implementation unfortunately throws an Error instead of returning false if the request was denied by the user. So I think we have to release here

Marukesu: sorry the delay, i didn't receive a notification about this one.

looking at the libportal code, cancelling in the dialog should throw a `IOError.CANCELLED`. using the ESC key or any other type of falling throw a `IOError.FAILED`.

so we should release if we catch a `IOError.CANCELLED`, otherwise we should only show the warning.

leolost2605: Sounds good, done!

Marukesu (approved):
LGTM.

PR #883, opened 2023-04-29
MessageListItem: Make Gravatar work again

### LINKED ISSUE #804: The app suspends when click an item in the MessageListBox
### What Happened?

The message in the inbox can't be open, if you click a message item, the `Mail` will suspend for a long time. 
Then, I try debug it using `gdb`, and something reads like the fellow:
```text
(io.elementary.mail:4188): Handy-WARNING **: 18:57:01.207: Failed to load icon: HTTP Error: Unable to connect: The connection timed out
```
After that, I search the code in this repository, and find out that maybe caused by `GravatarIcon`, as I find [this line](https://github.com/elementary/mail/blob/7241689418db9db6cdee1c35cc919faeec7d9e5a/src/MessageList/GravatarIcon.vala#L42).

I open `Chrome` and parse the url `https://secure.gravatar.com` into the address bar, and find out it can't be visit!

### Steps to Reproduce

1. block the internet access to `https://secure.gravatar.com`
2. open `io.elementary.mail`
3. click some messsage in the message list

### Expected Behavior

1. do not suspend the `GUI`
2. an option to switch off loading icons from `https://secure.gravatar.com` is better
3. only using icons from local is best

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
- Move loading Gravatars into a method in `MessageListItem` from its own class.
- Load Gravatars from a `http` address instead of `https` as `GLib.File.new_for_uri` seemingly doesn't support/has problems with https.
- Fix an issue that caused mail to freeze when a Gravatar wasn't found
- This also is a preparation for gtk4

Closes #804


danirabbit (approved):
Seems reasonable and works as expected. Nice simplification too!

PR #888, opened 2023-05-07
Unify trash/archive

### LINKED ISSUE #657: Move deleted mails to Trash folder

leolost2605:
- Unify trash and archive actions
- Add support for real trash folders
- Allow undoing of message archiving
- Prepare for a future move ability

This is based on #867 as it needs it for toasts and introduces a whole new system of expiring and undoing anyway.

Fix #657 but needs elementary/switchboard-plug-onlineaccounts#258 for this to really work

First part of a split of #881


danirabbit:
@leolost2605 can you resolve conflicts here?

danirabbit (approved):
This looks really good. Nice work with the new move action and making things very DRY and flexible. Delete and undo work as expected, afaict.

I can't confirm that I get an undo toast for Archive for some reason, but I also don't get one in master so :shrug:

PR #890, opened 2023-05-11
Always start InboxMonitor

leolost2605:
- Always start the InboxMonitor/the background daemon
- `--background` flag only signalizes that no window should be shown
- Fixes an issue that caused new messages to not get automatically fetched when mail was started without the `--background` flag
- Preparation for #882


src/Application.vala
```diff
@@ -137,45 +139,48 @@ public class Mail.Application : Gtk.Application {
...
 
-            main_window.show_all ();
+            main_window.present ();
+        } else {
+            run_in_background = false;
         }
 
-        main_window.present ();
+        if (first_run) {
+            first_run = false;
```
danirabbit, line 180: Would it be better to initialize the inbox monitor in `startup` instead of `activate`? I’m not sure about this `first_run` variable since it seems what we really care about is if the inbox monitor is running and the fact that it is started on first run is kind of incidental and we have a function for doing things once when the application is initialized

leolost2605: The thing is in the future we want to ask the flatpak portal for autstart and background permission everytime the app is started and to make the dialog transient we need the MainWindow already active, so we will probably need this `first_run` (maybe it should be called something else more similar to startup) thingy anyway. Or maybe there is a better way to do that? Is it possible to call activate in startup without any complications?

leolost2605: Another thing I forgot to mention was that we have to make sure that the `start ()` method from `Session` is called from the `MainWindow` first if mail is run without the background flag otherwise there might be some issues with the `WelcomeView` being briefly shown. An alternative here would be to move the entire `is_session_started`, `session_started` etc stuff into `Session` itself.

danirabbit (approved):
:shipit:

PR #891, opened 2023-05-17
Include attachments when forwarding

### LINKED ISSUE #729: Include Attachments when Forwarding an E-Mail
### Problem

When we forward an email with attachments, the attachments does not seem to be attached in the forwarded mail. While this is reasonable if we reply to an email, I think it would be better to append the attachment when we forward.

### Proposal

If the user forwards an email with attachments, also append the attachments to the forwarded mail.

### Prior Art (Optional)

_No response_

leolost2605:
- When forwarding get the attachments from the original message and include them in the new message
- Also make some margins ready for gtk4

Fixes #729


tintou (changes requested):
A few things to change and it looks alright to me

src/Composer.vala
```diff
@@ -682,13 +678,45 @@ public class Mail.Composer : Hdy.ApplicationWindow {
...
 
+    private async void get_attachments (Camel.DataWrapper message_content) {
+        if (message_content is Camel.Multipart) {
+            unowned var content = (Camel.Multipart)message_content;
+            for (uint i = 0; i < content.get_number (); i++) {
+                unowned var part = content.get_part (i);
+                unowned var field = part.get_mime_type_field ();
+                if (part.disposition == "attachment") {
+                    try {
+                        var file = File.new_for_path (Path.build_filename (Environment.get_tmp_dir (), part.get_filename ()));
```
tintou, line 700: I would create a temporary directory for each forwarding:

before the for loop:
```
GLib.File? tmp_dir = null;
```

here:
```suggestion
                        if (tmp_dir == null) {
                            tmp_dir = yield GLib.File.new_tmp_dir_async (null, GLib.Priority.DEFAULT, null);
                        }

                        var file = tmp_dir.get_child (part.get_filename ());
```
That way it is safe to forward several mails with the same filename.

leolost2605: `new_tmp_dir_async` is only available in GLib >= 2.74
So I used `GLib.File.new_for_path (Path.build_filename (Environment.get_tmp_dir (), Uuid.string_random ()))` with `make_directory` or is there an easier way?

zeebok: Can't you increase the version requirement since this is provided via flatpak?

leolost2605: Unfortunately Mail isn't provided as flatpak because evolution data server doesn't 100% work when flatpaked

zeebok (approved):
I don't know if you want @tintou to review again but it does work and nothing in the code stands out to me

src/Composer.vala
tintou: Sorry for the delay but you should use:
```suggestion
                            tmp_dir = GLib.File.new_for_path (GLib.DirUtils.make_tmp (null));
```
instead of `Uuid.string_random ()` which takes care of verifying that the directory doesn't exist previously.

tintou (approved):

PR #893, opened 2023-05-17
MessageList: Rewrite headers

danirabbit:
Makes expanded headers much more compact and makes it much easier to navigate conversations with changing subjects

* Use CSS to set padding and margins
* Show a subject header in the message list only when it changes instead in every message
* Slightly reduce the size of Avatars
* Don't show the "From:" label. Other fields like to, cc, bcc etc need disambiguation, but I think this is fairly clear that this first line is the sender
* Put all the actions in a row. This prevents things from moving around when messages have attachments etc
* Don't do clever overlapping with collapsed messages and always use the same header. A little less clever but things are always visible

## BEFORE
Screenshot from 2023-05-17 14 52 19

## AFTER
Screenshot from 2023-05-17 14 48 08

### DIFF
--- a/data/application.css
+++ b/data/application.css
@@ -7,8 +7,22 @@
         );
 }
 
-.card.collapsed {
-    margin-bottom: -64px;
+.message-list list {
+    padding: 0 1rem 0.25rem;
+}
+
+.message-list list > .h4 {
+    font-size: 1.25rem;
+    font-weight: 600;
+    padding: 0.75rem 0;
+}
+
+.message-list-item {
+    margin: 0.75rem 0;
+}
+
+.message-list-item.collapsed {
+    margin: 0.5rem 0;
 }
 
 .conversation-list-item {
--- a/src/MessageList/MessageList.vala
+++ b/src/MessageList/MessageList.vala
@@ -15,6 +15,7 @@ public class Mail.MessageList : Gtk.Box {
 
     construct {
         get_style_context ().add_class (Gtk.STYLE_CLASS_BACKGROUND);
+        get_style_context ().add_class ("message-list");
 
         var application_instance = (Gtk.Application) GLib.Application.get_default ();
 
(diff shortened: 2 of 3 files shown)


lenemter (commented):
I really don't like how much space this takes now. IMHO current design works great, why change it? Today I've seen people discussing this change in one of the elementary OS groups in Telegram and seems like people like compact design more.

However I like some changes: Hidden "From" label and actions in a row.

danirabbit:
@lenemter okay, I'll split that out! I would like to find a way to not repeat the subject line in every header and handle changing subjects more obviously but we can do things in stages and try to preserve the more compact design as well :)

danirabbit:
Hey @lenemter, I was able to get the overlapping back in there with some funky selectors, we still get separate headers for conversations where the subject changes (See the screenshot at the top I updated), and for single-subject conversations you can see the message headers are much more compact with no repeating subject line, and the datetime is centered vertically for the overlapped messages.

Screenshot from 2023-05-18 14 39 31
Screenshot from 2023-05-18 14 38 51

However, by overlapping again, we do lose the ability to see if a collapsed message was starred or has an attachment, so that seems problematic to me. I think from a perspective of doing the least hacky stuff, changing the header the least when collapsed, having the conversation list be usable at small window widths for tiling and mobile, and making sure all the information is visible for collapsed messages that overlapping is kind of a huge problem. Since it's in a scrolling list, I'm not sure using up vertical space is a huge issue.

lenemter:
> However, by overlapping again, we do lose the ability to see if a collapsed message was starred or has an attachment, so that seems problematic to me.

I didn't think about it, but it's a big issue now. I can't come up with an easy solution other than reverting this change :(

I really like the subject header but I noticed that this branch causes massive slowdowns when opening a thread on my end. Maybe because it needs to load a subject of every message?

lenemter:
I should have mentioned that I noticed slowdowns when deleting threads. When opening them performance matches one in `master`.

lenemter:
Sorry, just updated my Mail. Seems like the issue is present in `master` as well. Probably introduced by recent PRs.

leolost2605:
Probably with #888 then. Can you maybe open an issue with a clip or go into more detail what is slow? Like does it freeze or does it just not get trashed or similar stuff?

PR #895, opened 2023-05-20
Move capability

### LINKED ISSUE #800: Missing move ability

leolost2605:
- Allows moving messages between folders
- Fixes some criticals that were introduced with #888 

Any ideas, inspiration, suggestions etc. for a better UI are appreciated! Currently it looks like this:
Screenshot from 2023-05-20 15 24 27

Follow up on #888 
Second part of a split of #881
Fixes #800


src/MessageList/MessageList.vala
```diff
@@ -112,6 +114,14 @@ public class Mail.MessageList : Gtk.Box {
             tooltip_text = _("Mark Conversation")
         };
 
+        move_menu = new Menu ();
+
+        move_button = new Gtk.MenuButton () {
+            image = new Gtk.Image.from_icon_name ("folder", Gtk.IconSize.LARGE_TOOLBAR),
+            tooltip_text = _("Move Conversation to…"),
+            menu_model = move_menu
```
danirabbit, line 122:
```suggestion
            menu_model = move_menu,
            use_popover = false
```

Fixes padding issues with PopoverMenu in Gtk3

src/MessageList/MessageList.vala
```diff
@@ -295,6 +338,7 @@ public class Mail.MessageList : Gtk.Box {
 
     private void can_move_thread (bool enabled) {
         unowned var main_window = (Gtk.ApplicationWindow) get_toplevel ();
+        move_button.sensitive = enabled;
```
danirabbit, line 341: Probably should still mark the action here instead of touching button sensitivity directly in case we have other places in the UI where we have a move interface in the future or move action sensitivity is changed in another class etc

danirabbit (commented):
Just a couple things, but I think we can iterate on the UI in future branches. This is a great and functional start!

src/MessageList/MessageList.vala
leolost2605: Ah nice didn't know that!

src/MessageList/MessageList.vala
leolost2605: Got it. To still control the sensitivity I added `ACTON_MARK` (and renamed it to `ACTION_MODIFY`) to the button itself as it is used by the mark button for the same purpose.

leolost2605:
Should be ready again however I think it might be better to get #897 merged first

danirabbit (approved):

danirabbit:
@leolost2605 since #897 is on hold is it okay to merge this?

danirabbit (approved):
Still works as expected, nice job!

martinoroma88:
I cannot find the mentioned folder icon in the toolbar on top

danirabbit:
@martinoroma88 this was just merged into the development code base, so if you're in Early Access you should see it after updates, but it isn't yet generally available in the stable release

wout:
Will this also be the way to move emails to the spam folder?

leolost2605:
It definitely is a way, not sure whether a special move to spam button (similar to trash and archive) should be added?

wout:
I like the simplicity of it, but spam is a particular case.

It could also be an option under the flag button:
- Mark as unread
- Star
- Mark as spam
And then, when choosing the _Mark as spam_ option, the email is automatically moved to the spam folder.

Not sure if it makes sense because it would be slightly different behaviour than the two other options. But it's better than adding another dedicated button for spam alone, I think.

TomiOhl:
Totally agree with wout here, "Mark as spam" is an action in main clients including gmail's web interface, probably most people would search for this before thinking of moving the mail to the spam folder. It's just a UX thing based on people's familiarity

PR #897, opened 2023-05-20
Improve performance when moving messages

### LINKED ISSUE #896: Performance issues when trashing lots of conversations
### What Happened?

After https://github.com/elementary/mail/pull/888 trashing conversations became painfully slow. When you delete many conversations in a row mail freezes for a couple seconds.

### Steps to Reproduce

1. Open conversation
2. Delete it
3. Delete the next automatically opened conversation

### Expected Behavior

No freezes

### OS Version

7.x (Horus)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
- Improve performance when archiving, trashing and moving messages
- Queue move operations in order to improve performance
- Display a spinner when moving in order to not confuse users when they switch folders and the moved message isn't there yet
- Also includes a fix for some criticals introduced with #888 

Fixes #896


lenemter (changes requested):
Yes, it fixes performance issue for me too, but it doesn't address the issue with chaotic selection of messages after deleting and not updating the messages list correctly. For me this branch after deleting many conversations couldn't update the messages list. I press refresh button and nothing happens, and when restarting Mail UI it just can't load any messages.

lenemter:
Nevermind, seems like it actually can load messages, but it takes a long time.

leolost2605:
That's because when you switch folder (or also close the toast) the MoveHandler starts to move all of the queued messages at once. That's because otherwise after a message was moved it takes up to 5 seconds for it to appear in the folder it was moved to which might be confusing for the user if they switch to it sooner.

There might be multiple solutions:
1. keep it like this and accept long load times when many message were moved and the folder is switched (probably not the best)
2. only move messages all at once on folder switch if less than e.g. 5 messages are queued (probably the best)
3. Don't move messages all at once when switching folder might cause messages to just appear when you switched to the folder they were moved to before the 5 second timeout

Opinion on this i.e. 1 2 or 3 or an even better way? :)

leolost2605:
I went with option 3 for consistency reason and added a spinner in the bottom right corner.

lenemter (changes requested):
When deleting many conversations, some of them appear again in inbox after some time and disappear after refreshing folder. And 2 criticals are thrown when you try to open these buggy messages:
```
(io.elementary.mail:5071): camel-CRITICAL **: 16:31:03.576: camel_folder_summary_get_folder: assertion 'CAMEL_IS_FOLDER_SUMMARY (summary)' failed

(io.elementary.mail:5071): camel-CRITICAL **: 16:31:03.576: camel_folder_get_message: assertion 'CAMEL_IS_FOLDER (folder)' failed
```

danirabbit:
@lenemter im seeing that issue in master occasionally as well, I'm not sure it's related to changes in this branch

leolost2605:
@lenemter Tbh I'm running out of ideas here. I've completely rewritten this PR because I wanted to do that anyway to get some cleaner code so maybe that fixed it. However I have to say that I've never gotten your criticals neither before nor after so I don't know. Maybe they really were caused by something else as @danirabbit suggested because I think I've seen this errors somewhere else before too.
If not can you maybe give some more information like what kind of magnitude we're talking about when you say "many" messages (10, 50, 100?) or maybe it happens when e.g. all messages in a folder are deleted? Any additional info would be really appreciated!

Btw while I was at it I also added user facing errors some of them even being localized e.g. "No Archive folder is configured." or "Source folder is the destination folder." (With this one I'm sure there is a better way of saying it so suggestions are welcome as English is not my native language :)

danirabbit (approved):
I'm happy if @lenemter is happy :)

leolost2605:
I would still like to fine tune this a bit :)

leolost2605:
The last commit improves performance/fixes issues when moving lots of conversations in particular conversations with lots of messages.
It also fixes some issues like when a conversation had lots of messages it was possible to start multiple move operations and sometimes the first message was then selected.
@lenemter I hope your issue had to do with this but I once again couldn't 100% replicate it. Still your description shares similarities with the issues I had because of this so this is my best bet :)

It's ready again but should be tested thoroughly.

lenemter (approved):
Works very nicely. Thanks for your work :heart:

PR #899, opened 2023-05-22
Fix an issue that causes mail to never connect to the network

leolost2605:
- Fix an issue that was introduced with #871 that causes mail to not connect to the network because `NetworkMonitor.network_available` is for some reason `false` in the beginning. To do this we set `Session.online` to true by default (which btw is also the default if it's not set at all but I think it's better to keep explicitly setting it to avoid confusion) 
- Add some debug messages
- Alternatively we could also use GLib.NetworkMonitor which reports a change in the beginning so a connection is established because of the signal connection (might be better but I would leave this to a different PR to get the issue at hand fixed ASAP)

I'm so sorry, I had this commit locally but forgot to push 😬


danirabbit (approved):
Ah dunno how I missed this! It seemed to work okay here

PR #900, opened 2023-05-22
Add support for signatures

### LINKED ISSUE #616: Email account signature
## What Happened
Cannot find where to set email account signature

## Expected Behavior
Should be able to set an email signature for each email account setup

## Steps to Reproduce
Hard to try explain how to reproduce, as I don't know what to click on in order to find where to set the signature ;-)

### Logs
No logs available

## Platform Information
Elementary OS 6 Daily Builds with latest updates

leolost2605:
- First iteration of signature support
- Requires #894 

TODO:
- [x] A lot of cleanup
- [x] Iterate on UI with new discussions in #616 

Edit Signatures Dialog:
Screenshot from 2023-06-01 18 01 08

The signature icon in the composer needs improvements because I've got no idea how to create good SVGs and couldn't find a good one with an appropriate license online.

Fixes #616


danirabbit:
@leolost2605 could you either merge or rebase this to resolve conflicts please? :)

leolost2605:
@danirabbit done.

lenemter:
I have some concerns about the UI part of this PR.

1. "Title" header has weird bottom margin
2. We probably can move the "Remove" button to the list itself (some elementary apps use this pattern) and get rid of the second action bar.
3. The margins of "Create signature" are very small, IMO it should be like in Printers settings
image
image

4. I think placeholder has some room for improvements. What about

#### No Signatures
Add new signature using button bellow.

---

Asking @danirabbit for her opinion.

zeebok:
I am not Dani but I agree @lenemter I think the margins needs adjusting and the remove button moved.

danirabbit:
@lenemter @zeebok pushed some fixes for spacing, placeholder, etc. I do want to keep the "delete" button in the actual content area instead of in the sidebar to make it really clear that it's tied to this signature

Screenshot from 2023-07-19 11 18 46

But thoughts on actionbar versus box here? Also if @micahilbery has opinions

Screenshot from 2023-07-19 11 17 48

zeebok:
I think if the intent is to keep the delete button with the active signature I think I like the latter more. Helps it really feel tied together as one concept.

lenemter:
@danirabbit "Title" header still has large bottom spacing because it's packed into the headerbar.

> But thoughts on actionbar versus box here?

I like box more here :)

leolost2605:
@danirabbit thank you for taking this over! The last commit only removes some redundant parent setting since it's now set by the application.

Regarding the bottom margin mentioned by @lenemter, it's only indirectly related to the headerbar but more because we add an additional `top_margin = 2` to the title entry because otherwise the highlight around it when focused would be overlapped (?) by the headerbar and therefore not visible and I wasn't sure what to do about it :)

danirabbit:
Yeah that can just be removed in Gtk4. And we can pack a WindowControls to get even more control over spacing there as well. So I’m not super worried about it since it’s kind of a temporary quirk of Gtk 3

zeebok:
Things look good but I do have a couple UX questions. If I add a signature to the email content I see:

1. It creates a section of email that I have to either have to tab to or click on to edit it. Is this intended or standard behavior for signatures? I don't really use them so I don't know if that is common but it isn't intuitive.
2. If I try to remove the signature by erasing it in the content area, that content area still exists and I can't get back to the body content without doing what I described in 1. I can only remove it by picking None for the signature. Does it need to be cleaned up if it is empty or will that happen at another point in the code?

leolost2605:
@zeebok AFAIK geary does it and I think some other clients like windows mail. Also it was a really easy way to make it reliably work. However I agree it currently is unintutive if you don't really know what's going on so if we keep it we probably should add some outlines to make clear what is currently edited (main body, signature or quote).

Also if anyone wants to, you can checkout the branch [signatures-intuitive](https://github.com/elementary/mail/tree/signatures-intuitive) and tell me which version you prefer. There I've tried my best (ignore the weird cursor placing for now :)) to make it more intuitive but that needs to be tested thoroughly because otherwise it might lead to some data loss if the user accidently manages to write everything in the signature div because it wasn't easy distinguishable and then changes the signature to none.

zeebok:
@leolost2605 Yeah that branch definitely felt more how I would expect it to behave. I think it is fine to merge this and upgrade the feel separately where we can have a focused changelist and conversation

zeebok (approved):

micahilbery:
> @lenemter @zeebok pushed some fixes for spacing, placeholder, etc. I do want to keep the "delete" button in the actual content area instead of in the sidebar to make it really clear that it's tied to this signature
> 
> Screenshot from 2023-07-19 11 18 46
> 
> But thoughts on actionbar versus box here? Also if @micahilbery has opinions
> 
> Screenshot from 2023-07-19 11 17 48

hmm 🤔 I like how clean the bottom action bar looks but it feels a little disconnected from the content it effects. Idk both are fine with me (super helpful I know 😜 )

zeebok:
@danirabbit and @leolost2605 Anything else you want to discuss or change before this gets merged?

leolost2605:
As far as I'm concerned we can merge this. I'll open another PR for further discussions around behavior.

PR #902, opened 2023-05-23
Application: force use of Portals

danirabbit:
This makes Mail use Portals even when run as a host app. This means we use the FileChooser portal instead of the Gtk FileChooser widget


Marukesu:
This possibily will break notifications when running in the host.

danirabbit:
afaict, notifications are still working with this branch

src/Application.vala
```diff
@@ -25,6 +25,8 @@ public class Mail.Application : Gtk.Application {
     }
 
     construct {
+        Environment.set_variable ("GTK_USE_PORTAL", "1", true);
```
Marukesu, line 28: i think we shouldn't try to override the environment if already set externally. also a FIXME or TODO comment saying "remove this once we have Gtk.FileDialog" would be good, so that we don't forget to remove this workaround later.

danirabbit: Ah does Gtk.FileDialog always use portals even on the host?

We're currently doing this in Code and System Settings for the same reason: https://github.com/search?q=org%3Aelementary+Environment.set_variable+%28%22GTK_USE_PORTAL%22%2C+%221%22%2C+true%29%3B&type=code

Marukesu: yes, they try the portal first and if that fails, they use a fallback Gtk.FileChooser.

Marukesu (approved):
Works as expected.

PR #905, opened 2023-05-31
Move menu to FoldersListView

danirabbit:
Screenshot from 2023-05-31 11 40 20

* Makes room for the undo toast so it's less likely to cover buttons
* Settings related to the whole app/accounts aren't in the deepest pane, this will probably matter more for responsive/mobile but it makes them less logically connected to the messages list

### DIFF
--- a/src/FoldersView/FoldersListView.vala
+++ b/src/FoldersView/FoldersListView.vala
@@ -134,9 +134,9 @@ public class Mail.FoldersListView : Gtk.Grid {
 
         account_settings_menuitem.clicked.connect (() => {
             try {
-                AppInfo.launch_default_for_uri ("settings://accounts/online", null);
+                Gtk.show_uri_on_window ((Gtk.Window) get_toplevel (), "settings://accounts/online", Gdk.CURRENT_TIME);
             } catch (Error e) {
-                warning ("Failed to open account settings: %s", e.message);
+                critical ("Failed to open account settings: %s", e.message);
             }
         });
     }
--- a/src/FoldersView/FoldersListView.vala
+++ b/src/FoldersView/FoldersListView.vala
@@ -56,11 +56,47 @@ public class Mail.FoldersListView : Gtk.Grid {
         var scrolled_window = new Gtk.ScrolledWindow (null, null);
         scrolled_window.add (source_list);
 
+        var load_images_menuitem = new Granite.SwitchModelButton (_("Always Show Remote Images"));
+
+        var account_settings_menuitem = new Gtk.ModelButton () {
+            text = _("Account Settings…")
+        };
+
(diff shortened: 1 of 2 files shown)


src/FoldersView/FoldersListView.vala
```diff
@@ -93,6 +128,17 @@ public class Mail.FoldersListView : Gtk.Grid {
                 settings.set ("selected-folder", "(ss)", "GROUPED", grouped_folder_item.name);
             }
         });
+
+        var settings = new GLib.Settings ("io.elementary.mail");
+        settings.bind ("always-load-remote-images", load_images_menuitem, "active", SettingsBindFlags.DEFAULT);
+
+        account_settings_menuitem.clicked.connect (() => {
+            try {
+                Gtk.show_uri_on_window ((Gtk.Window) get_toplevel (), "settings://accounts/online", Gdk.CURRENT_TIME);
```
Marukesu, line 137: i would try to use a valid timestamp here.
```suggestion
                Gtk.show_uri_on_window ((Gtk.Window) get_toplevel (), "settings://accounts/online", Gtk.get_current_event_time ());
```

src/FoldersView/FoldersListView.vala
```diff
@@ -93,6 +128,17 @@ public class Mail.FoldersListView : Gtk.Grid {
...
+            } catch (Error e) {
+                critical ("Failed to open account settings: %s", e.message);
```
Marukesu, line 139: i believe we should show a dialog here, this was a direct user interaction, we shouldn't silently fail.

src/FoldersView/FoldersListView.vala
```diff
@@ -56,11 +56,46 @@ public class Mail.FoldersListView : Gtk.Grid {
...
+        var account_settings_menuitem = new Gtk.ModelButton () {
+            text = _("Account Settings…")
+        };
+
+        var app_menu_separator = new Gtk.Separator (Gtk.Orientation.HORIZONTAL) {
+            margin_bottom = 3,
+            margin_top = 3
+        };
+
+        var app_menu_box = new Gtk.Box (VERTICAL, 0) {
```
Marukesu, line 70: shouldn't this be Gtk.Orientation.VERTICAL rather than VERTICAL? to be inline with the rest of the file.
```suggestion
        var app_menu_box = new Gtk.Box (Gtk.Orientation.VERTICAL, 0) {
```

danirabbit: I don't think we have anything in the style guide about it. I just found out we could do this recently. IMO, it makes things more legible, but I'm not picky about it

src/FoldersView/FoldersListView.vala
danirabbit: Added an error dialog

Screenshot from 2023-06-01 13 19 27

Marukesu (approved):
LGTM.

PR #907, opened 2023-06-02
Add support for inline images in the composer

leolost2605:
- Add support for inline images in the composer using cid + mimepart. CID + mimepart seems to be the most widely used technique for local inline images (alternatives would e.g. be a base64 encoded image directly inside the html body).
- Also update some margins for gtk4


zeebok:
Code looks good, I will give this a test tonight!

leolost2605:
Ok so the last commit removes the image when the attachment is removed. I'll probably do the same for the other way round.

Also I thought maybe we shouldn't show the attachment at all?

zeebok:
> Also I thought maybe we shouldn't show the attachment at all?

If that can be done without too much circumvention of code, yeah that might be the cleanest experience. Maybe @danirabbit or @elementary/ux have a strong reason to have the attachment visible?

zeebok:
But the fix you provided to match the inline and the attachment works like a charm!

leolost2605:
No attachment is shown now (i just set `no_show_all` on inline attachments :))
Also only images can now be selected in the filechooser

zeebok (approved):

PR #908, opened 2023-06-07
Move the composer editor toolbar into its own class

leolost2605:
- Moves the editing toolbar of the composer (bold, italic, underline, strike through, clear format, link) into its own class
- This is done so that it can be used in multiple places (to be specific it will be useful for creating a signature)


leolost2605:
Done 😅

danirabbit (approved):
Everything works as expected :tada:

PR #910, opened 2023-06-16
Sort special folders to top

### LINKED ISSUE #903: Show special folders at the top level of the account sidebar tree
### Problem

Some email providers (like GMail) have special folders buried within other folders instead of at the top level in the sidebar. This means by default many of the special folders are hidden:

Screenshot from 2023-05-23 14 11 40

### Proposal

We should always show those special folders at the top level of the tree instead, even though that's not where they really are on the server:

Screenshot from 2023-05-23 14 12 52

### Prior Art (Optional)

This is what iOS Mail, Geary, and Mailspring all do

leolost2605:
- Sorts special folder to the top of the accounts folder tree
- There they currently are sorted in this order:
    - INBOX
    - DRAFTS
    - OUTBOX
    - SENT
    - ARCHIVE
    - TRASH
    - JUNK

Fixes #903

### DIFF
--- a/src/FoldersView/AccountSourceItem.vala
+++ b/src/FoldersView/AccountSourceItem.vala
@@ -123,7 +123,7 @@ public class Mail.AccountSourceItem : Mail.SourceList.ExpandableItem, Mail.Sourc
             }
 
             if (folder_item.is_special_folder) {
-                this.add (folder_item);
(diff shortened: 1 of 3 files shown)


danirabbit (approved):
Works great. Thank you! Huge QoL boost!

PR #911, opened 2023-06-16
Allow renaming of folders

leolost2605:
- Allow folder renaming
- Uses the SourceList editor
- Also does some cleanup like moving the `refresh_folder` method into the `FolderSourceItems`
- It uses toasts to notify about errors (e.g. when a folder with the same name already exists) because i think they are less obtrusive than dialogs but that can be changed of course
- The `start_editing_item` method in SourceList exists according to documentation in the original Granite.SourceList but was probably removed when making it internal so I added it back


src/FoldersView/FolderSourceItem.vala
```diff
@@ -98,5 +117,76 @@ public class Mail.FolderSourceItem : Mail.SourceList.ExpandableItem {
...
+        Camel.FolderInfo? folder_info = null;
+        try {
+            folder_info = yield offlinestore.get_folder_info (new_full_name, FAST, GLib.Priority.DEFAULT, cancellable);
+        } catch (Error e) {
+            warning (e.message);
+        }
+
+        if (null != folder_info) {
+            if (name == old_name) {
+                notify["name"].connect (() => { name = old_name; });
```
zeebok, line 167: If the new name is the same as the old, meaning a person decides to not rename, shouldn't it just return? In testing, it feels a bit odd to get an error from backing out of the rename

leolost2605: Yeah that's definitely unwanted. Last commit fixes that and some more issues you might have been having

zeebok (approved):
This works as advertised! Thanks for the work!

lenemter (approved):
I noticed that toasts can overflow the window, but I guess that's out of scope of this branch. Otherwise works as expected.

image

danirabbit:
@lenemter ah yeah for error cases we should probably throw a dialog anyways since toasts are so ephemeral

PR #914, opened 2023-06-26
Control add portals to depends

danirabbit:
Depend on new enough portals for creating autostart using the background portal

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -23,7 +23,7 @@ Homepage: http://github.com/elementary/mail
 
 Package: io.elementary.mail
 Architecture: any
-Depends: ${misc:Depends}, ${shlibs:Depends}
+Depends: io.elementary.portals (>=7.1.0), ${misc:Depends}, ${shlibs:Depends}
 Description: Email client
  Mail is an email client built for elementary's Pantheon desktop environment. It
  allows you to read and send email with a simple, modern interface.


debian/control
```diff
@@ -23,7 +23,7 @@ Homepage: http://github.com/elementary/mail
 
 Package: io.elementary.mail
 Architecture: any
-Depends: ${misc:Depends}, ${shlibs:Depends}
+Depends: io.elementary.portals (>=7.1.0), ${misc:Depends}, ${shlibs:Depends}
```
Marukesu, line 26: we actually don't depend in the backend, only in the frontend. so a `xdg-desktop-portals (>=1.4.0)` is really the only thing mail depends.

danirabbit: I think we want to make sure that newer portals is pulled in before installing this though right? Otherwise if this update is installed and portals is not, users would be left without an autostart right?

Marukesu: Right, there's a `xdg-desktop-portal-backend` virtual package, but I don't think the portal has it in they provides.

danirabbit: Sorry, I meant, not only do we need portals to be installed but we specifically need this version of our portal backend for this feature to continue working right? That's why I'm pulling in this version of portals is to make sure someone doesn't accidentally install this update without updating portals and then run into a feature regression

lenemter (approved):

PR #915, opened 2023-06-29
Show an InfoBar if attachments include an ICS file

leolost2605:
- Show an InfoBar that tells the user the message comes with a calendar event and allows the user to open it right away
- Also when opening attachments without saving them their temp file will now be in a separate directory within the temp dir so that it can keep its original name to avoid strange names like `AZAY61-calendar.ics`
- Suggestion for better phrasing are always appreciated :)


zeebok:
Code looks good. Not sure what is the cause of the issue but if I use the button to open in calendar without Calendar already running, it doesn't seem to work. I'll see the icon for the app but no window ever appears. Using the button when calendar is running works like a charm.

zeebok (approved):
The bug I saw seems to be Calendar's issue so this is approved

PR #916, opened 2023-07-02
Improve move menu and make it searchable

### LINKED ISSUE #904: Search move menu
### Problem

If you have a lot of folders/tags it can be a hassle to find the right folder when you're doing a move

### Proposal

Add a search to the move menu so I can more easily filter down to the right folder

### Prior Art (Optional)

Mailspring:
Screenshot from 2023-05-30 10 50 03

leolost2605:
It gets moved into a separate class because it might come in handy later in other places (e.g. moving folders etc.)

Screenshot:
Screenshot from 2023-07-02 14 12 22

Fixes #904


zeebok (approved):
Code seems reasonable and it works like a charm

PR #921, opened 2023-07-13
MessageList: use mail-move icon

danirabbit:


leolost2605 (approved):
Where does that come from/How can you find it? It's neither listed in the Icon Browser nor shipped with mail or is it? Is there somewhere else one can look to find more icons? Would be really helpful to know :)

danirabbit:
@leolost2605 it’s in the icons repo! I need to double check, maybe it’s not in the fd.o spec. If it’s not then it maybe be a good idea to either move it here or make sure we document it in the icon browser

PR #923, opened 2023-07-13
FolderPopover: increase white space, natural height, search placeholder

danirabbit:
Also activate first row when pressing enter from search entry

Screenshot from 2023-07-13 10 54 32

Screenshot from 2023-07-13 10 54 40

Screenshot from 2023-07-13 11 06 21


leolost2605 (changes requested):
Looks really good, two things to note though. Is there something one can do about the popover changing width when the placeholder gets show with a long search term (maybe use multiple lines or something like that?). Also I wouldn't expect the first row to get activated when pressing enter in the search entry but maybe that's just me :)

danirabbit:
@leolost2605 Removed the search activate, and I also removed repeating the search term in the placeholder. When wrapping it looks like the scrolledwindow height isn't updated for some reason so this seems kinda janky:

Screenshot from 2023-07-15 10 14 36

PR #926, opened 2023-07-17
MainWindow: throw dialog for errors

danirabbit:
Fix overflow issue pointed out in #911 by making sure we use dialogs for errors instead of toasts

Screenshot from 2023-07-17 11 26 04
Screenshot from 2023-07-17 11 18 00
Screenshot from 2023-07-17 11 17 46
Screenshot from 2023-07-17 11 48 16


leolost2605 (approved):
Looks good!

PR #927, opened 2023-07-20
Remember recipients

### LINKED ISSUE #743: Autocomplete recipients using known addresses
### Problem

Seems fairly self-explanatory. I have contacts. I'm currently not really given any mechanism to manage those contacts.

### Proposal

I'd like to be able to manage contacts using some form of app. Maybe not integrated into Mail though since contacts could be used in the calendar invites, for example.

Also, Mail should be able to autocomplete my recipients in the To, CC, and BCC fields using aforementioned contacts.

### Prior Art (Optional)

_No response_

leolost2605:
- When sending a message add all recipients to the primary addressbook which defaults to the `eds:system-address-book`
- This allows these addresses to be autocompleted (I've seen some more or less recent issues about autocomplete not working however it does for me, if somebody has still issues it would be good to know :))
- Also as they are currently added to the primary addressbook they will show up in e.g. gnome contacts under the "Local Address Book"
- We also respect the history settings (I think that's right here, or is it?)
- This has nothing to do with any online account contacts that come with configured email accounts. They are still not recognized and/or edited

Fixes #743


src/Backend/ContactManager.vala
```diff
@@ -67,6 +70,63 @@ public class Mail.ContactManager : GLib.Object {
...
 
+    public async void remember_mail_address (string address, string? name) {
+        name = name ?? address;
+
+        if (!Granite.Services.System.history_is_enabled ()) {
+            debug ("History disabled, don't remember email addresses.");
+            return;
+        }
+
+        if (address in email_addresses) {
```
zeebok, line 81: Is this totally necessary? add_persona_from_details already checks for existing details for us, and this list isn't utilized anywhere else.

leolost2605: I know it says in the docs it does but with my testing it still creates additional entries that for example show up in GNOME Contacts... maybe I'm doing something wrong?

zeebok:
This one really got away from me. I am going to try and test this by the end of this coming weekend. Besides the one comment, which your code may be necessary, this code seems fine. Would be good to see this get merged and closed!

src/Backend/ContactManager.vala
zeebok: I have been looking at the docs more and I wonder if what we should be doing is seeing if an individual exists with the details in question, and if they do use that as the parent in the `add_persona_from_details` call? That way if it is the same person but different email, the contact gets appended to instead of a new contact with the same name.

zeebok (approved):
Finally got to dig into this more. I can confirm that on the current release in 7.1 it doesn't seem to remember recipients and this fix does the trick. Thanks for this!

PR #930, opened 2023-07-22
Swipe to archive and trash

### LINKED ISSUE #551: Support for Multi-touch track pad gestures
## Proposal
Provide multi touch track pad gestures for archiving, deleting messages or other performing other actions from the messages list.

## Prior Art
The Mac OS Mail app does this.

https://user-images.githubusercontent.com/68198116/110114277-6376ba80-7dda-11eb-8881-63916036e578.mp4

Also, almost every mobile mail app has swiping features.

leolost2605:
Fixes #551

Swipe right to archive, swipe left to trash.

If somebody knows CSS the backgrounds of the action parts need improvements (I'm not very familiar with CSS and didn't find something like a `@destructive_color` so idk :)) because it definitely can't be merged as is (I had suggested action for archive and destructive action for trash in mind but am open to suggestions). Also some more UX questions might be debatable like what actions should it be (maybe mark read instead of archive as I think Apple does) or whether mouse drag should be disabled.

This also includes some GTK4 prep and bug fixes as it moves the `create_context_menu` method into the `ConversationListItem` which brings the advantage that we don't rely on `get_row_at_y` anymore which isn't available in GTK4's ListView and already was a bit buggy especially if the list was scrolled down and then with this PR the coordinates reported by `button_press_event` were wrong entirely probably because of the carousel 🤷. (For some reason using the `action_name` on the MenuItems didn't work so I had to activate the actions manually, any ideas why that is?)


leolost2605:
@lenemter thanks for the PR!

danirabbit:
Proposed gresourcing icons including a symbolic here: https://github.com/elementary/mail/pull/933

data/ConversationListItem.css
```diff
@@ -0,0 +1,41 @@
...
+}
+
+archive-affordance.right {
+    border-top-left-radius: 6px;
+    border-bottom-left-radius: 6px;
+}
+
+archive-affordance image,
+archive-affordance label {
+    color: mix(@theme_fg_color, @success_color, 0.6);
```
danirabbit, line 39: iOS mail uses purple which I think is quite nice:

Screenshot from 2023-07-22 13 35 46

```suggestion
    background-color: mix(@theme_bg_color, @GRAPE_500, 0.3);
    padding: 12px;
}

archive-affordance.left {
    border-top-right-radius: 6px;
    border-bottom-right-radius: 6px;
}

archive-affordance.right {
    border-top-left-radius: 6px;
    border-bottom-left-radius: 6px;
}

archive-affordance image,
archive-affordance label {
    color: mix(@theme_fg_color, @GRAPE_500, 0.6);
```

leolost2605: It does look nice :)

src/ConversationList/ConversationListItem.vala
```diff
@@ -83,13 +95,67 @@ public class Mail.ConversationListItem : VirtualizingListBoxRow {
...
+        carousel.add (trash_affordance);
+        carousel.scroll_to (grid);
+
         get_style_context ().add_class ("conversation-list-item");
-        add (grid);
+        child = carousel;
 
         show_all ();
+
+        button_release_event.connect ((e) => {
```
danirabbit, line 117: While we’re here can we use an event controller? https://valadoc.org/gtk+-3.0/Gtk.GestureSingle.html

lenemter:
It would be really nice if these gestures were customizable e.g. move instead of archive.

danirabbit:
@lenemter I agree it would be nice to allow selecting different actions, but I think it could wait for another branch so this doens't get stuck in review for a long time

src/ConversationList/ConversationListItem.vala
leolost2605: TBH I've never managed to get Controllers to work in GTK3 :(

When I do this in the `ConversationListItem`:

```vala
        var gesture = new Gtk.GestureMultiPress (this) {
            button = Gdk.BUTTON_SECONDARY,
            propagation_phase = BUBBLE
        };
``` 

I never get a signal. Any ideas why that is?

lenemter: In Gtk3 widgets don't contain references to controllers, so you need to save your `gesture` to a field. Just add `private Gtk.GestureMultiPress gesture;` in class.

leolost2605: Oh that makes sense then, probably over read it somewhere :(
Thank you, it's fixed now!

leolost2605:
This also fixes the issue that the mark menu in the headerbar didn't get updated when a message for example switched from read to unread.

danirabbit (approved):
Nice work! I think this is working as expected and code improvements look great too

PR #935, opened 2023-07-25
Update metainfo

leolost2605:
A lot was already added in #920 this only adds some more small things that were added since then and fixes a little issue.

I probably should've done this with every PR and will try to do that from now on :)
Also as I've successfully avoided updating metainfo until now if I did something wrong please tell me.


zeebok:
Don't forget about the new, shiny signature support you added!

PR #937, opened 2023-07-25
ConversationListItem: set menu attach widget

danirabbit:
Set menu attach widgets so that `action_name` works as expected

while here, use `child` and avoid long lines


leolost2605 (approved):
`attach_widget`... I'll have to remember that :)

danirabbit:
Yeah I think it’s because Gtk.Menu is technically a separate window so it’s not aware of its parent window unless we attach it to a widget in that tree

PR #939, opened 2023-07-31
Refresh folder when selecting it

leolost2605:
Fixes an issue that when adding a new account the messages in its folders wouldn't show up right away but first after the folder was at least once manually refreshed.

### DIFF
--- a/src/ConversationList/ConversationList.vala
+++ b/src/ConversationList/ConversationList.vala
@@ -247,6 +247,9 @@ public class Mail.ConversationList : Gtk.Box {
             cancellable.cancel ();
         }
 
+        var cancellable = new GLib.Cancellable ();
+        this.cancellable = cancellable;
+
         conversation_focused (null);
         conversation_selected (null);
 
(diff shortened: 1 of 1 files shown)


src/ConversationList/ConversationList.vala
```diff
@@ -311,19 +312,28 @@ public class Mail.ConversationList : Gtk.Box {
...
-                    folder.parent_store.display_name,
-                    e.message);
+                    if (e is IOError.CANCELLED) {
+                        break;
+                    } else {
+                        warning ("Error fetching messages for '%s' from '%s': %s",
+                        folder.display_name,
+                        folder.parent_store.display_name,
+                        e.message);
+                    }
```
lenemter, line 336: You can simplify it like this:

```
} catch (IOError.CANCELLED e) {
    ...
} catch (Error e) {
    ...
}
```

leolost2605: Oh nice thanks!

leolost2605:
I'm noticing during development in general that sometimes on cold start mail's folder list doesn't load. I'm not 100% sure but to me it seems this occurs more often with this branch so I'll convert to draft for now to do some more testing

PR #951, opened 2023-09-07
MessageListItem: Fix resizing

leolost2605:
Fixes some annoying resizing of the window or the paned position when the window is small and sender/recipient addresses or the subject include a long word by using the wrap mode `WORD_CHAR`.

### DIFF
--- a/src/MessageList/AttachmentButton.vala
+++ b/src/MessageList/AttachmentButton.vala
@@ -90,7 +90,9 @@ public class AttachmentButton : Gtk.FlowBoxChild {
         };
 
         name_label = new Gtk.Label (mime_part.get_filename ()) {
-            xalign = 0
+            xalign = 0,
+            wrap = true,
+            wrap_mode = WORD_CHAR
         };
 
         size_label = new Gtk.Label (null) {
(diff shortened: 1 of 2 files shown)


jeremypw (changes requested):
This fixes the issue for some messages, but not if they have an attachment with a long title (which does not wrap). In that case the window still resizes.

jeremypw:
This now fixes the attachment issue. The only thing I don't like is that the minimum width of the address fields is only about 4 chars so you can get very tall thin fields.  Not sure what the best solution is to this.   Maybe the date field should also be responsive and/or a minimum width/height applied to the address fields?

leolost2605:
Now a min width of 10 (debatable i'm open to suggestions) characters is applied and the datetime label is ellipsized in the middle (also debatable). I've got no special preference here so tell me what you think :)

(And cc is now also wrapped with word_char, i forgot that previously)

jeremypw:
I find wrapping the date with `wrap_mode = WORD` looks better (in English locale at least) and all the info remains visible - what do you think? With the date shrinking the min width of the address is less important as the overall minimum width is limited by the header bar anyway, but 10 is OK.

leolost2605:
@jeremypw done :)

PR #955, opened 2023-09-09
Don't disable delete accelerators when the ConversationList looses keyboard focus

### LINKED ISSUE #952: Make delete/backspace key work with focus on email
### Problem

Currently, the `delete`/`backspace` buttons only work when the focus is on the email in the list column (e.g. it's highlighted in colour instead of gray). Sometimes, the list item loses focus after loading the email. Other times, it loses focus after clicking on the email content.

### Proposal

Make the `delete`/`backspace`  buttons work even when the focus is not on the list column.

### Prior Art (Optional)

_No response_

leolost2605:
With undoing working very reliable now, I don't think there is a reason to disable the keyboard accelerators for trashing a message when the ConversationList looses keyboard focus. It only is annoying when clicking once within the message and then being unable to use `Backspace` or `Delete` for trashing it.

Fixes #952

### DIFF
--- a/src/ConversationList/ConversationList.vala
+++ b/src/ConversationList/ConversationList.vala
@@ -165,22 +165,6 @@ public class Mail.ConversationList : Gtk.Box {
 
         search_entry.search_changed.connect (() => load_folder.begin (folder_info_per_account));
 
-        // Disable delete accelerators when the conversation list box loses keyboard focus,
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
I think this may have originally been done because of the inline composer, but as far as I can tell any of the remaining inputs work as expected: you can delete text without accidentally deleting an email

PR #956, opened 2023-09-09
Use the actual recipients of a message as sender when replying or forwarding

leolost2605:
This fixes an issue that occurred when replying to a message received by an alias. Then the pre-selected sender would be the main address.

### DIFF
--- a/src/Composer.vala
+++ b/src/Composer.vala
@@ -554,6 +554,8 @@ public class Mail.Composer : Hdy.ApplicationWindow {
             bool found = false;
 
             if (type != DRAFT) {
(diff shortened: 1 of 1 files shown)


danirabbit (commented):
I think I understand what this is doing, but can you add a comment in the code about what/why this does so that someone doesn't come along later and remove it thinking it's superfluous :sweat_smile:

leolost2605:
I've added some comments and I hope they are understandable but tell me if they are unclear :)

danirabbit (approved):
Looks good, thank you!

PR #961, opened 2023-09-11
Improve aliases design

lenemter:
My attempt to improve the design.

image


leolost2605 (approved):
Thanks so much, I like it a lot!

Issue #957, opened 2023-09-10
Can't download attachment from email


wout:
### What Happened?

I have one email with an attachment that I can't download. It's a PDF in a fairly simple email.

![image](https://github.com/elementary/mail/assets/107324/2f2af5df-b73b-4c18-b9d7-4f4638e2b139)

### Steps to Reproduce

1. Open email
2. There's no download button/file icon

### Expected Behavior

A file icon should appear to be able to download the file.

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

### Hardware Info

```
         eeeeeeeeeeeeeeeee            wout@Framey 
      eeeeeeeeeeeeeeeeeeeeeee         ----------- 
    eeeee  eeeeeeeeeeee   eeeee       OS: elementary OS 7 Horus x86_64 
  eeee   eeeee       eee     eeee     Host: Laptop (12th Gen Intel Core) A8 
 eeee   eeee          eee     eeee    Kernel: 6.2.0-32-generic 
eee    eee            eee       eee   Uptime: 3 hours, 27 mins 
eee   eee            eee        eee   Packages: 1956 (dpkg), 69 (flatpak) 
ee    eee           eeee       eeee   Shell: bash 5.1.16 
ee    eee         eeeee      eeeeee   Resolution: 3000x2000, 3840x2160 
ee    eee       eeeee      eeeee ee   DE: Pantheon 
eee   eeee   eeeeee      eeeee  eee   WM: Mutter(Gala) 
eee    eeeeeeeeee     eeeeee    eee   Theme: io.elementary.stylesheet.mint [GTK3] 
 eeeeeeeeeeeeeeeeeeeeeeee    eeeee    Icons: elementary [GTK3] 
  eeeeeeee eeeeeeeeeeee      eeee     Terminal: io.elementary.t 
    eeeee                 eeeee       CPU: 12th Gen Intel i7-1280P (20) @ 4.700GHz 
      eeeeeee         eeeeeee         GPU: Intel Alder Lake-P 
         eeeeeeeeeeeeeeeee            Memory: 9149MiB / 64003MiB 
```

Issue #950, opened 2023-09-05
Many mails flash to white after opening


wout:
### What Happened?

It happens very often that after opening a mail it shows for a split-second, but then turns completely white. It also happens with emails from senders where it didn't happen on previous mails (e.g. Collosal).

![image](https://github.com/elementary/mail/assets/107324/564a916a-43af-42a2-bd64-20e49eb69d76)

Closing and re-opening mail does not help, nor does it change which email turns white and which one doesn't. That's always consistent.

[2023-09-05-18:02:38.391749663.mkv.tar.gz](https://github.com/elementary/mail/files/12526764/2023-09-05-18.02.38.391749663.mkv.tar.gz)

Printing the email does work.

### Steps to Reproduce

1. Click on mail
2. It loads and shows
3. It flashes completely to white

### Expected Behavior

Mail content should not flash to white.

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
This is from the syslog, but it also and happens with mails that do load so I'm not sure if it's useful:

Sep  5 20:07:51 Framey io.elementary.m[786301]: MessageListItem.vala:509: Not Found
Sep  5 20:07:51 Framey xdg-desktop-por[2656]: Realtime error: Could not map pid: Could not determine pid namespace: Could not find instance-id in process's /.flatpak-info
```

### Hardware Info

```
         eeeeeeeeeeeeeeeee            wout@Framey 
      eeeeeeeeeeeeeeeeeeeeeee         ----------- 
    eeeee  eeeeeeeeeeee   eeeee       OS: elementary OS 7 Horus x86_64 
  eeee   eeeee       eee     eeee     Host: Laptop (12th Gen Intel Core) A8 
 eeee   eeee          eee     eeee    Kernel: 6.2.0-31-generic 
eee    eee            eee       eee   Uptime: 1 day, 7 hours, 27 mins 
eee   eee            eee        eee   Packages: 1922 (dpkg), 69 (flatpak) 
ee    eee           eeee       eeee   Shell: bash 5.1.16 
ee    eee         eeeee      eeeeee   Resolution: 3000x2000 
ee    eee       eeeee      eeeee ee   DE: Pantheon 
eee   eeee   eeeeee      eeeee  eee   WM: Mutter(Gala) 
eee    eeeeeeeeee     eeeeee    eee   Theme: io.elementary.stylesheet.mint [GTK3] 
[...]

PR #963, opened 2023-09-14
Allow donating a message to elementary

### LINKED ISSUE #409: Donate an email for debugging
Some way to easily donate a misbehaving email to us so we could take a look and debug it would be useful. 

At the very least, providing the full original contents of the mail so the user can manually file an issue would be good. More interestingly, automatically uploading it somewhere (with proper consent, of course) could be more frictionless.

leolost2605:
This would be very useful to investigate issues like #957 or #950 where normal forwarding doesn't work as that just takes the html and builds an entirely new message whereas this only modifies sender and recipient of the original message and sends it as is. For now it gets sent to a throwaway address of mine but it probably would be good to create an account (maybe even an elementary one?) that all contributors have access to (make credentials available in slack?) and change it to that before merging this. 
Any feedback regarding design, phrasing or whether this might be problematic in a way would be very appreciated! :)

Below every message:
image

When the link is clicked:
Screenshot from 2023-09-14 18 31 24

Fixes #409

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -24,7 +24,7 @@ public class Mail.MainWindow : Hdy.ApplicationWindow {
 
     private FoldersListView folders_list_view;
     private Granite.Widgets.Toast move_toast;
-    private Granite.Widgets.Toast error_toast;
+    private Granite.Widgets.Toast info_toast;
     private ConversationList conversation_list;
     private MessageList message_list;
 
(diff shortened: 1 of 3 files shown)


leolost2605:
> Would it be possible to hide that in 3 dot menu for each message? If that line would be added to every message then it would be quite jarring.

I agree that this might be annoying when added below every message. My issue with the three dot menu was that I didn't find a good short phrase that would describe what this does (e.g. _Donate_ seemed a bit vague and confusing to me), any ideas? 

> TBH, something even better would be an ability to save a complete email into a file (outlook has that option).

That sure is possible if we want it and could be implemented additionally. Still I would prefer the way this PR works over having to download the source and then create an issue with that source because the latter just adds a lot more friction. Also with access to the account for donation messages maybe being somewhat limited to elementary contributors this would maybe preserve a little bit more privacy with not everyone being able to just read the messages and mail address.

wout:
Love the idea!

> (e.g. _Donate_ seemed a bit vague and confusing to me), any ideas?

What about **Send feedback...** to bring it in line with the Feedback app?

jlnr:
> TBH, something even better would be an ability to save a complete email into a file (outlook has that option).

macOS Mail supports message drag and drop as EML files, which can then be compressed and attached to GitHub issues. That also works to attach one message to another one within the mail app. I find that quite elegant, and it also makes it easier to audit the information I'm sending as EML is a plain-text format.

PR #967, opened 2023-10-24
AliasesDialog: screen reader fixes, design tweaks

danirabbit:
Alias:
* Use a vertical form layout with header labels to match the design of other forms
* Make sure the screen reader knows how to describe entries by setting mnemonic_widget
* Placeholder text in forms should contain example content, not explanation text or directions
* Image buttons should have tooltip text (screenreader reads this too)

AliasDialog:
* Subclass Granite.Dialog since we're not doing anything that requires a Hdy.Window
* Dialogs should have a transient parent so they are centered on the parent etc
* Make sure screen reader can read the add button
* Use built-in method for dialog buttons instead of adding window controls manually

### DIFF
--- a/src/Dialogs/AliasDialog/Alias.vala
+++ b/src/Dialogs/AliasDialog/Alias.vala
@@ -45,18 +45,16 @@ public class Mail.Alias : Gtk.ListBoxRow {
             xalign = 0
         };
 
-        var edit_name_label = new Gtk.Label (_("Name:")) {
-            halign = END
-        };
-
         var name_entry = new Gtk.Entry () {
+            margin_start = 12,
+            margin_end = 12,
             text = alias_name,
-            placeholder_text = _("Enter name…")
+            placeholder_text = _("John Doe")
         };
         name_entry.bind_property ("text", this, "alias-name", DEFAULT);
 
-        var edit_address_label = new Gtk.Label (_("Address:")) {
-            halign = END
+        var edit_name_label = new Granite.HeaderLabel (_("Name")) {
+            mnemonic_widget = name_entry
         };
 
         Regex? regex = null;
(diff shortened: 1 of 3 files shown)


src/Dialogs/AliasDialog/AliasDialog.vala
```diff
@@ -92,27 +89,23 @@ public class Mail.AliasDialog : Hdy.ApplicationWindow {
...
+        overlay.show_all ();
 
         title = _("Aliases");
         default_height = 300;
         default_width = 500;
-        add (overlay);
-        show_all ();
+        get_content_area ().add (overlay);
+        this.add_button (_("Close"), Gtk.ResponseType.CLOSE);
         present ();
```
leolost2605, line 109: We probably should call `present ()` after creating the dialog and not in `construct` otherwise I get this warning:
```
(io.elementary.mail:64822): Gtk-WARNING **: 17:56:20.812: gtk_window_set_titlebar() called on a realized window
```

src/FoldersView/AccountSourceItem.vala
```diff
@@ -152,7 +152,11 @@ public class Mail.AccountSourceItem : Mail.SourceList.ExpandableItem, Mail.Sourc
         var menu = new Gtk.Menu ();
 
         var alias_item = new Gtk.MenuItem.with_label (_("Edit Aliases…"));
-        alias_item.activate.connect (() => new AliasDialog (account.service.uid));
+        alias_item.activate.connect (() => {
+            new AliasDialog (account.service.uid) {
+                transient_for = ((Gtk.Application) GLib.Application.get_default ()).active_window
+            };
```
leolost2605, line 158:
```suggestion
            }.present ();
```

Here :)

leolost2605 (changes requested):
Thank you looks really nice, I only have some minor comments :)

src/Dialogs/AliasDialog/Alias.vala
```diff
@@ -165,6 +168,7 @@ public class Mail.Alias : Gtk.ListBoxRow {
             });
 
             start_delete ();
+            start_delete ();
```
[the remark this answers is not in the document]
danirabbit: Oops, dunno how that happened

src/Dialogs/AliasDialog/AliasDialog.vala
danirabbit: Nice catch

leolost2605 (approved):
Very nice 🎉

PR #979, opened 2024-02-08
Add apparmor profile

leolost2605:
On OS 8 mail crashes by default with these errors when opening an html message:
```
bwrap: Creating new namespace failed: Permission denied

** (io.elementary.mail:8681): ERROR **: 20:09:48.735: Failed to fully launch dbus-proxy: Child process exited with code 1
```

This seems to be caused by changes in how ubuntu configures bubblewrap with apps now requiring apparmor profiles to access it ?
I'm not really sure what's actually going on behind the scenes but as per [this launchpad bug](https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2046844) and my own testing creating an apparmor profile for mail fixes the issue.

### DIFF
--- a/data/io.elementary.mail.apparmor
+++ b/data/io.elementary.mail.apparmor
@@ -1,0 +1,9 @@
+abi <abi/4.0>,
+include <tunables/global>
+
+profile io.elementary.mail /usr/bin/io.elementary.mail flags=(unconfined) {
+  userns,
+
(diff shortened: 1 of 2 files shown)


danirabbit:
It looks like one of the affected projects marked on the launchpad report is bubblewrap. I wonder if we should be waiting for a fix in bubblewrap?

danirabbit (approved):

PR #986, opened 2024-05-18
update list of dependencies in README

eighthave:
In order to get this building on Debian/bookworm, I had to install these.


danirabbit (approved):
Thanks for updating this list!

PR #1023, opened 2025-06-19
Adapt to libcamel API changes in 3.57.1

### LINKED ISSUE #1042: Mail fails to build with latest evolution-data-server package
### What Happened?

The evolution-data-server package on my machine got updated (from 3.56.2 to 3.58.3). The package manager detected a lib dependency on the old version and tried to rebuild Mail. This failed due to various errors (EDS API seems to have changed).

### Steps to Reproduce

Build mail 8.0.1 while having evolution-data-server 3.58.3 installed (3.56.2 was still working)

### Expected Behavior

Build succeeds and results in a working binary (like it does when building against eds 3.56)

### OS Version

Other Linux

### OS Architecture

amd64 (on most hardwares)

### Session Type

Classic Session (X11)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

mcrha:
The upcoming evolution-data-server 3.57.1 development release (planned for the end of the next week) contains API changes in libcamel, which are incompatible with the previous versions. Bump the libcamel version requirement, update the camel-1.2.vapi file and modify the sources accordingly.

@tintou would you mind to have a look, please? I do not speak vala, this change is more or less blindly done, into the way the compiler succeeds, but whether the type casts I added are correct or could be done better or some others are valid I do not know. Note for example of the `get_uids()`, which changed to `dup_uids()`, which returns "transfer container" array now, where it used to return "transfer full" array with a special function to free it.


danirabbit (changes requested):
This breaks builds for the current stable release of elementary OS, so this needs to be done by adding a build flag so it can continue to build for both versions. And ideally we need to add a container to CI where this version does succeed. Like maybe a newer Fedora container?

mcrha:
I just rebased the branch against the current master. Hope it helps.

PR #1054, opened 2026-07-22
ConversationListItem: use GLib.Menu

danirabbit:


jeremypw (approved):
So much more efficient!

Still uses Gtk.Menu but I guess you are working towards replacing that as well

PR #1055, opened 2026-07-22
SourceList: move interfaces to their own files

danirabbit:
Move public interfaces to their own file as per code style


jeremypw (approved):
This looks like the same SourceList as is used by Code, which I found too difficult to port.  I am working on a suitable Gtk4 replacement.  Is it worth putting it in Granite-7 as it is used in (at least) two projects?

Anyway this PR does no harm

danirabbit:
@jeremypw i wanted to try to see if I could pull it apart and maybe get closer to a smaller diff 😅

I think previously we had removed the source list because it turned out that we had a bunch of custom sidebar implementations with different requirements. Most of the sidebar implementations don't actually need a tree for example.

Maybe we can figure out if there are common components or interfaces or styles we can find a way to share without making a single overly complex widget?

PR #1060, opened 2026-07-27
FolderList: use GLib.Menu for main menu

danirabbit:
ModelButton is sealed in GTK4. Just use a menu model here


ryonakano (approved):
Confirmed the menu (now changed from a popover) shown as before and each menuitem works as expected.

PR #1069, opened 2026-07-28
AccountItemModel: Use GLib.Menu

danirabbit:
GTK4 prep


ryonakano (approved):
Confirmed the menu is shown and works as expected.

PR #1078, opened 2026-08-06
Composer: use ListStore for attachments

danirabbit:
Can't use `get_children ()` in GTK4, so use a ListStore here

### DIFF
--- a/src/Composer.vala
+++ b/src/Composer.vala
@@ -23,12 +23,12 @@ public class Mail.Composer : Hdy.ApplicationWindow {
 
     private bool discard_draft = false;
     private Camel.MessageInfo? ancestor_message_info = null;
+    private ListStore attachment_list;
 
     private WebView web_view;
     private Gtk.Entry to_val;
     private Gtk.Entry cc_val;
     private Gtk.Entry bcc_val;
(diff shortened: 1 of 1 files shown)


src/Composer.vala
```diff
@@ -216,10 +218,11 @@ public class Mail.Composer : Hdy.ApplicationWindow {
         var editor_toolbar = new EditorToolbar (web_view);
         editor_toolbar.add (image);
 
-        attachment_box = new Gtk.FlowBox () {
+        var attachment_box = new Gtk.FlowBox () {
             homogeneous = true,
             selection_mode = Gtk.SelectionMode.NONE
         };
+        attachment_box.bind_model (attachment_list, (obj) => { return (Attachment) obj; });
```
leolost2605:
```suggestion
        attachment_box.bind_model (attachment_list, (obj) => (Attachment) obj);
```
You can just do this and it will treat it as a return statement. Saves a few braces and make it a bit more legible IMO

leolost2605 (approved):
Only a minor nit apart from that LGTM

src/Composer.vala
danirabbit: Oh sick. Looks good ty

PR #1079, opened 2026-08-06
MessageList: Use ListStore for messages

danirabbit:
Can't use `get_children` in GTK4


jeremypw (approved):
Code looks equivalent and no regressions found.

PR #1080, opened 2026-08-06
FolderPopover: use ListStore

danirabbit:
can't use `get_children` in GTK4


jeremypw (approved):
Code looks equivalent and no regressions found.

PR #1081, opened 2026-08-06
SignatureDialog: Use ListStore

danirabbit:
can't use get_children in GTK4. We can technically iterate over the listbox still, but it's better to iterate over data


src/Dialogs/SignatureDialog/SignatureDialog.vala
```diff
@@ -55,12 +56,15 @@ public class Mail.SignatureDialog : Hdy.ApplicationWindow {
...
+        signature_list = new ListStore (typeof (Signature));
+
+        signature_list_box = new Gtk.ListBox () {
             vexpand = true,
             selection_mode = BROWSE
         };
-        signature_list.set_filter_func ((Gtk.ListBoxFilterFunc)filter_func);
-        signature_list.set_placeholder (placeholder);
+        signature_list_box.bind_model (signature_list, (obj) => (Signature) obj);
+        signature_list_box.set_filter_func ((Gtk.ListBoxFilterFunc)filter_func);
```
jeremypw, line 66: Isn't this ignored when bound to a model?  I get the terminal message `GtkListBox with a model will ignore sort and filter functions`

jeremypw (changes requested):
I think the filter function will have to be reimplemented a different way, otherwise looking good.

danirabbit:
@jeremypw in GTK4 we can use a filter list model. Does this not work for you?

jeremypw:
> @jeremypw in GTK4 we can use a filter list model. Does this not work for you?

In Gtk4 you still cannot use the `set_filter_func` function of the ListBox but yes, you could wrap the model in a Gtk.FilterListModel.  I haven't looked deeply into it but the code looks weird anyway - the filter is dependent on the presence/absence of a timout.  I would have thought there is a better way of doing it.  Can't we just remove the item from the model when it starts to get deleted?

danirabbit:
@jeremypw I think the idea is we don't want to actually delete it right away because we want to be able to undo. So we're just hiding it from the UI with the filter until we do the actual delete.

I believe you can still use filters and listmodels together in GTK3 even though it warns about it, but I don't think there's any way to really fix the warning in GTK3 :woman_shrugging:

jeremypw:
@danirabbit You can certainly leave the code in place if you wish - but I don't think the filter will work will it?  That may not matter.  If it will stay for a significant time and does not work maybe a comment is appropriate?

I now realise the warning is given regardless of whether any sort or filter func is applied.

danirabbit:
@jeremypw it still works in my testing? Does it not work for you?

danirabbit:
Also beware of https://github.com/elementary/mail/issues/1086 when testing

jeremypw:
@danirabbit Hmm, the documentation only says it is "incompatible" not that it doesn't work I guess. Just like you can still add ListBoxRows directly to a ListBox even if it is bound to a model.  But then the model does not match the box.  I'll check this more thoroughly to see whether there are any drawbacks to applying a filter_func in this case.

jeremypw:
Well, as far as I can see, signatures are never deleted from the model - they are just filtered out by the listbox which seems weird. Can you see a reason for this? Why not just delete them from the model?   Maybe its because of the "Undo" facility? But that seems unnecessary as signatures are easy to create?

jeremypw:
But as you say the deleted filter does seem to work, at least in Gtk3.  May not in Gtk4 though.

danirabbit:
@jeremypw it should be removed from the model when the window is closed

jeremypw:
Ah, yes there is a 5 second delay before the signature is *actually* deleted.  The undo function just stops the timeout.  I guess it is safe enough to leave it as it is for now.

jeremypw (approved):

PR #1082, opened 2026-08-06
AliasDialog: use ListStore

danirabbit:
Can't use `get_children` in GTK4, so use a ListStore


src/Dialogs/AliasDialog/AliasDialog.vala
```diff
@@ -51,23 +52,26 @@ public class Mail.AliasDialog : Granite.Dialog {
...
+
+        list_box = new Gtk.ListBox () {
             vexpand = true,
             hexpand = true,
             selection_mode = NONE
         };
-        list.set_filter_func ((Gtk.ListBoxFilterFunc) filter_func);
-        list.set_placeholder (placeholder);
+        list_box.bind_model (alias_list, (obj) => (Alias) obj);
+        list_box.set_filter_func ((Gtk.ListBoxFilterFunc) filter_func);
```
jeremypw, line 63: I think filter functions are ignored when there is a bound model.

src/Dialogs/AliasDialog/AliasDialog.vala
```diff
@@ -51,23 +52,26 @@ public class Mail.AliasDialog : Granite.Dialog {
...
-            child = list,
+            child = list_box,
             hscrollbar_policy = NEVER
         };
 
         var add_button_label = new Gtk.Label (_("Add Alias…"));
 
         var add_box = new Gtk.Box (HORIZONTAL, 0);
-        add_box.add (new Gtk.Image.from_icon_name ("list-add-symbolic", Gtk.IconSize.SMALL_TOOLBAR));
+        add_box.add (new Gtk.Image.from_icon_name ("list_box-add-symbolic", Gtk.IconSize.SMALL_TOOLBAR));
```
jeremypw: I don't think you meant to change this?  The icon does not exist.

jeremypw (changes requested):
Couple of inline comments.  Also clicking on the add button does nothing.  Investigating.

src/Dialogs/AliasDialog/AliasDialog.vala
```diff
@@ -173,19 +177,21 @@ public class Mail.AliasDialog : Granite.Dialog {
...
+            uint pos = -1;
+            if (alias_list.find (alias, out pos)) {
+                alias_list.remove (pos);
+            };
+
             aliases.remove (alias.address);
             write_aliases ();
         });
-
-        list.add (alias);
```
jeremypw, line 188: You need to replace this with somthing!

src/Dialogs/AliasDialog/AliasDialog.vala
danirabbit: Oops yeah thanks for catching that 😅

danirabbit:
@jeremypw oops sorry about that. Should work now

jeremypw (approved):
Looks good now and seems to work.

PR #1084, opened 2026-08-07
SourceList: GTK4 prep

danirabbit:
* `install_style_property` and `style_get` are gone in GTK4
* Replace constants removed in GTK4
* remove unnecessary `Granite.STYLE_CLASS_SOURCE_LIST` 
* remove unnecessary `stock_size`
* remove `public override bool popup_menu ()` which is gone in GTK4
* `align` is `FILL` by default, only needs `vexpand`
* remove dead CSS
* clean up code style

Screenshot from 2026-08-07 09 56 07


jeremypw:
Looks like you are going to port SourceList as is to Gtk4.  In that case, I'll copy your work intoCode when you have finished if that's OK.

danirabbit:
@jeremypw going to attempt to at least! And yes absolutely if it works out please :)

jeremypw:
Probably the only place SourceList is fully used now is in Code's projects sidebar.  Photos has something similar but its own code. It seems under-used in Mail tbh.

jeremypw (approved):
Appearance looks pretty much the same and code sane.

PR #1085, opened 2026-08-09
CellRendererExpander: hard code arrow size

danirabbit:
`style_get` doesn't exist in GTK4, so hardcode this to the smallest icon size which is 16


jeremypw (approved):
Makes sense.

